PR #263, opened 2025-07-30
fixes [Bug]: Padded task numbers do not work

varadan13:


MrLesk:
Hi @varadan13 
while your fix does indeed fixes the issue, I just noticed that we should not generate a task ID in the server component. It should be responsibility of the Core. I will add a proper refactor on top of your PR and do a cleanup of the duplicated logic

PR #484, opened 2026-01-16
Add migration to normalize lowercase task IDs

### LINKED ISSUE #483: Add migration to normalize legacy lowercase task IDs
## Problem

Task files created manually, imported, or created by older versions may have lowercase IDs (e.g., `id: task-3`) instead of the expected uppercase format (`id: TASK-3`).

This causes reorder operations to fail with errors like:
```
Task TASK-3 not found while reordering
```

The `reorderTask` function normalizes input IDs to uppercase, but then compares against the raw ID from the file using strict equality.

## Root Cause

The `saveTask()` function correctly normalizes IDs to uppercase on save (line 191-198 in `operations.ts`):
```typescript
// Normalize task ID and parentTaskId to uppercase before serialization
const normalizedTask = {
    ...task,
    id: taskId,  // Normalized via normalizeId()
```

But files created outside this flow (manual creation, imports, older versions) may have lowercase IDs.

## Proposed Solution

Add a one-time migration (similar to `prefix-migration.ts`) that:

1. Scans task files in `tasks/` and `completed/` directories
2. Checks if the `id` field is lowercase
3. Rewrites files with normalized uppercase IDs
4. Runs automatically on startup when needed

This maintains the intended case-sensitive behavior while fixing legacy data.

## Alternative Considered

Making ID comparisons case-insensitive via `taskIdsEqual()` - rejected because it works around the problem rather than fixing the root cause.

saada:
## Summary

Fixes #483

Adds an automatic migration that normalizes lowercase task IDs to uppercase. This fixes reorder failures for legacy files created manually, imported, or by older versions.

## Problem

Task files with lowercase IDs (e.g., `id: task-3`) cause reorder operations to fail:
```
Task TASK-3 not found while reordering
```

The `reorderTask` function normalizes input IDs to uppercase, but compares against raw IDs from files using strict equality.

## Solution

Add a migration (following the existing `prefix-migration.ts` pattern) that:

1. Scans `tasks/` and `completed/` directories on startup
2. Detects files with lowercase IDs (e.g., `id: task-3`)
3. Rewrites them with uppercase IDs (e.g., `id: TASK-3`)

This maintains the intended case-sensitive behavior while automatically fixing legacy data.

## Changes

- `src/core/id-case-migration.ts` - New migration module
- `src/core/backlog.ts` - Integrate migration into `ensureConfigMigrated()`

## Test plan

1. Create task files with lowercase IDs (e.g., `id: task-3`)
2. Start backlog browser
3. Migration runs automatically, normalizing IDs to uppercase
4. Reorder operations now succeed

### DIFF
--- a/src/test/acceptance-criteria.test.ts
+++ b/src/test/acceptance-criteria.test.ts
@@ -555,7 +555,7 @@ describe("AcceptanceCriteriaManager unit tests", () => {
 			expect(result.exitCode).toBe(0);
 
 			// Parse task ID from output
-			const taskId = result.stdout.toString().match(/Created task (task-\d+)/)?.[1];
+			const taskId = result.stdout.toString().match(/Created task (TASK-\d+)/)?.[1];
 			expect(taskId).toBeTruthy();
 
 			// Verify ACs were created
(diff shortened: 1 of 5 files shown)


MrLesk:
Hey @saada . Thank you very much for this PR. The case that you found is something that I tried to cover and the current behavior is a bug. I would not go with a migration solution that might remain in the code forever but rather try to enforce the id normalization I wrote recently.

PR #499, opened 2026-01-22
BACK-361 - Add label-based filtering to TUI and web UI task list views

MrLesk:
## Summary
- rebase BACK-361 onto current main and drop stale formatter-only changes
- keep the existing TUI/Web label-filtering implementation already present on main
- show selected labels in the TUI filter header button using the shared concise label summary helper
- disable blessed tag parsing for popup buttons so label text renders literally

## Testing
- bun test src/test/label-filter.test.ts src/test/filter-header-navigation.test.ts
- bunx tsc --noEmit
- bun run check .
- bun test


MrLesk:
@codex review

MrLesk:
@codex review

PR #555, opened 2026-03-07
BACK-442 - Prevent DoD defaults config injection

MrLesk:
### Motivation
- Follow-up to BACK-394 / PR #542 to ensure MCP Definition of Done defaults cannot inject additional config keys when saved.
- Current config serialization already writes DoD defaults with JSON string escaping, so quoted and multiline values should remain valid user content instead of being rejected.

### Description
- Implements `BACK-442 - Prevent DoD defaults config injection` as regression coverage for the existing safe serialization path.
- Adds MCP coverage for a quoted, multiline DoD default containing an `onStatusChange`-looking payload.
- Verifies the payload round-trips as a DoD default and does not create or modify `onStatusChange` in loaded config.
- Keeps apostrophe-bearing defaults accepted.

### Testing
- `bun test src/test/mcp-definition-of-done-defaults.test.ts`
- `bun run check .`
- `bunx tsc --noEmit`

### Related
- Task: BACK-442
- Follow-up to BACK-394 / PR #542

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69ac757bb4c48333bbd2b53adb3167c1)

### DIFF
--- a/src/test/mcp-definition-of-done-defaults.test.ts
+++ b/src/test/mcp-definition-of-done-defaults.test.ts
@@ -158,4 +158,44 @@ describe("MCP Definition of Done default tools", () => {
 		const reloaded = await loadConfigOrThrow(server);
 		expect(reloaded.definitionOfDone).toEqual(["Run unit, integration, and e2e tests"]);
 	});
+
+	it("round-trips quoted and multiline DoD defaults without injecting config keys", async () => {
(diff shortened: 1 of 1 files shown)


MrLesk:
Alex's Agent: This PR has passing checks and a valid-looking title, but I could not see a matching Backlog task included in the PR; current `main` has `BACK-394` with a different task title. Please add/update/link the Backlog task or retitle to the actual task before merge.

PR #563, opened 2026-03-14
BACK-402 - Support project-local custom backlog directories

### LINKED ISSUE #215: [Feature]: Proposal - Use .backlog directory to align with standard conventions
Hey there! First off, I'm really enjoying using and contributing to [backlog.md](code-assist-path:/workspaces/Backlog.md/backlog.md). It's a fantastic tool.

As I've integrated it more into my projects, I've noticed that it creates a backlog/ directory in the project root. While this works perfectly fine, many other developer tools use "dot-directories" (like .vscode/, .github/, .idea/) to keep project-specific tooling files neatly tucked away.

### The Proposal

I'd like to propose an enhancement to make the backlog/ directory name configurable. This would allow users who prefer the "dot-directory" convention to use .backlog/ instead, making the tool feel even more integrated and keeping the project root cleaner, without changing the current default behavior for existing users.

### Implementation Idea

I've looked into the code and it seems the main challenge is the tool needs to read [config.yml](code-assist-path:/workspaces/Backlog.md/backlog/config.yml) to know the name of the backlog directory, but it needs to know the directory name to find the config file in the first place.

A potential solution to this would be:

1. **Make discovery smarter**: Instead of assuming the directory is named backlog, the tool could search for [config.yml](code-assist-path:/workspaces/Backlog.md/backlog/config.yml) in a predefined list of locations, for example ['.backlog', 'backlog']. This would make the tool flexible and maintain backward compatibility for all existing projects.
2. **Add an option to init**: The init command could be updated with a --directory flag, allowing a user to run backlog init --directory .backlog. The default behavior of backlog init would remain unchanged.
3. **Store the setting**: The chosen directory name would be saved in the backlog_directory field within the [config.yml](code-assist-path:/workspaces/Backlog.md/backlog/config.yml) file, so all subsequent commands know where to find the files.

This seems like an architectural improvement that would enhance the user experience by offering more flexibility. I'm happy to help with a PR if this is a direction you'd be open to.

What are your thoughts on this?

### LINKED ISSUE #334: [Feature]: Custom backlog dir & multiple backlogs
If I can have a custom backlog folder then I can do something like this:

backlogs/
    backlog-1/
    backlog-2/

..then run it like `backlog backlog-1 browser`

This is useful to split up the project into multiple perspectives and is a stopgap for not having swimlanes.

Currently, the workaround is to store backlogs somewhere else and symlink them into the project.

MrLesk:
## Summary
- use deterministic project-root discovery via `backlog.config.yml`, `backlog/`, and `.backlog/`
- make root `backlog.config.yml` the canonical config file when present, with optional `backlog_directory` for custom project-relative backlog folders
- update CLI/web init and filesystem config resolution to use the root-config model, including explicit non-interactive folder/config-location selection

## Validation
- `bunx tsc --noEmit`
- `bun run check .`
- `bun test`

Closes #215
Closes #334

### DIFF
--- a/backlog/tasks/back-402 - Support-project-local-custom-backlog-directories.md
+++ b/backlog/tasks/back-402 - Support-project-local-custom-backlog-directories.md
@@ -5,7 +5,7 @@ status: In Progress
 assignee:
   - '@codex'
 created_date: '2026-03-14 19:55'
-updated_date: '2026-03-14 23:26'
+updated_date: '2026-03-15 10:30'
 labels: []
 dependencies: []
 references:
@@ -65,6 +65,8 @@ Fixing root detection so walking up from nested custom backlog folders requires 
 Fixing resolver fallback so an invalid placeholder root backlog.config.yml does not suppress an otherwise valid folder-local backlog config project.
 
 Fixing built-in folder precedence so `backlog/` only wins over `.backlog/` when it has a config marker; otherwise resolution falls back to the built-in folder that actually contains backlog config.
+
+Fixing shared init validation so backlogDirectorySource and backlogDirectory must agree when both are provided, preventing callers from smuggling a custom path under a built-in source.
 <!-- SECTION:NOTES:END -->
 
 ## Final Summary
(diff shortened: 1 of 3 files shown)


MrLesk:
@codex check again

Issue #192, opened 2025-07-10
[Bug]: CLAUDE.md not updated on init


searls:
The init suggests it's going to paste the AGENT_GUIDELINES into CLAUDE.md during init

I ran: 

```
mkdir foo; cd foo; touch CLAUDE.md; git init; git add .; git commit -m 'lol'; backlog init example
```

Then selected CLAUDE.md interactively and hit enter and init exited status code 0.

```
cat CLAUDE.md
```

Yielded an empty file. 

Had to dig through the repo to extract agent guidelines manually

MrLesk:
@searls I'm trying to reproduce the problem on backlog 1.1.8 but when initializing backlog with the sequence you provided I get the correct file.
Could you please let me know OS and maybe check also the other agents files to see if they have the same issue? Also try to see if when running backlog init when there is already a claude.md file, it appends the text correctly. You can run backlog init twice

MrLesk:
Alex's Agent: The original repro steps are clear, but the init and agent-instruction flows have changed substantially since this was reported. I recommend retesting on the latest release with the same CLAUDE.md selection flow, plus backlog agents --update-instructions. If the file is populated now, this can close as stale/fixed; if not, it should become a current init/agent-instructions bug task.

MrLesk:
Alex's Agent: I retested this against the current built CLI and the current source flow.

Reproduction checked:
1. Created a clean git repo with an existing empty `CLAUDE.md`, committed it, then ran `backlog init example`.
2. In the init wizard, accepted the default storage/config/task-prefix prompts, selected CLI instructions, toggled `CLAUDE.md` with Space, accepted with Enter, and skipped advanced settings.
3. Result: init exited 0, reported `Updated: CLAUDE.md`, and `CLAUDE.md` was populated with the managed `<!-- BACKLOG.MD GUIDELINES START -->` block including the `backlog instructions overview` guidance.

Additional checks:
- Pressing Enter at the init instruction-file prompt without toggling a file now retries with `Please select at least one agent instruction file before continuing.` It does not finish silently.
- In an initialized project with an empty `CLAUDE.md`, `backlog agents --update-instructions` with `CLAUDE.md` selected exits 0 and writes the same managed guidance block.

Automated validation:
- `bun test src/test/agent-selection.test.ts src/test/agent-instructions.test.ts src/test/cli-agents.test.ts src/test/cli.test.ts` -> 123 pass, 0 fail.

[...]

PR #575, opened 2026-03-22
BACK-440 - Handle cancel in agents update prompt

Luc0-0:
## Problem
Issue #192 reports confusing behavior around instruction updates and init flows. In the interactive `backlog agents --update-instructions` flow, cancelling the multiselect prompt was handled the same as an empty selection path.

## Root Cause
The code converted a cancelled prompt result directly into an empty `files` array, so command flow continued instead of treating cancel as an explicit abort.

## Solution
Added explicit cancel handling in `src/cli.ts`:
- detect `clack.isCancel(selected)`
- log a clear cancellation message
- return early from the command action

Then process selected files only for non-cancelled responses.

## Related Tasks
- Closes BACK-440
- Refs #192

## Task Checklist

## Validation
- Reproduced current-main behavior with `/usr/bin/expect` against a real pseudo-terminal: Escape at the multiselect exits via `No files selected for update.`
- Verified this PR branch with `/usr/bin/expect`: Escape prints `Agent instruction update cancelled.` and does not create instruction files.
- Verified Enter with no selection still prints `No files selected for update.`
- Verified Space+Enter still updates `CLAUDE.md`.
- `bun test src/test/cli-agents.test.ts`
- `bunx tsc --noEmit`
- `bun run check .`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -3125,12 +3125,12 @@ agentsCmd
 					{ label: "Copilot (GitHub Copilot)", value: ".github/copilot-instructions.md" },
 				],
 			});
-			const files: AgentInstructionFile[] = clack.isCancel(selected)
-				? []
-				: Array.isArray(selected)
-					? (selected as AgentInstructionFile[])
-					: [];
+			if (clack.isCancel(selected)) {
+				clack.log.info("Agent instruction update cancelled.");
+				return;
+			}
 
+			const files: AgentInstructionFile[] = Array.isArray(selected) ? (selected as AgentInstructionFile[]) : [];
 			if (files.length > 0) {
 				// Get autoCommit setting from config
 				const config = await core.filesystem.loadConfig();


Luc0-0:
Quick follow-up on this one whenever you have time.

This change treats cancel in the update-instructions prompt as an explicit stop, so the command exits cleanly instead of continuing as if nothing was selected.

I kept the diff focused to src/cli.ts and re-ran the related agent selection path locally. Happy to tweak behavior or add extra coverage if you want.

MrLesk:
Alex's Agent: Thanks for the focused fix. Please add or link the corresponding Backlog task and retitle the PR to `BACK-<id> - <task title>` so it matches the project PR workflow.

MrLesk:
Alex's Agent: I created and pushed BACK-440, updated this PR title/body, rebased the branch on current main, and verified the fix with real pseudo-terminal runs. Base/main reproduces the problem by printing 'No files selected for update.' on Escape; this PR prints 'Agent instruction update cancelled.' and leaves instruction files unchanged. Empty Enter and Space+Enter selection paths still behave correctly. Local checks passed: bun test src/test/cli-agents.test.ts, bunx tsc --noEmit, and bun run check .. GitHub Actions are green and the PR is clean/mergeable. I am leaving it unmerged until the required Codex review reaction appears on head dfdf843.

Luc0-0:
Thanks for picking this up and doing the BACK-440 task creation, rebase, and PTY verification — really appreciate it. Standing by for the Codex review and happy to address anything else if needed.

MrLesk:
Alex's Agent: @codex review

I re-checked this against current main. BACK-440 still appears to be the right task ID for this focused cancel-path fix. I updated the PR body from `Closes #192` to `Refs #192` because #192 is broader than this change and should not be auto-closed by merging this PR.

PR #612, opened 2026-04-26
BACK-395 - Web: prevent stale SPA loads by disabling index.html caching

### LINKED ISSUE #611: [Bug]: Backlog browser can get stuck on stale chunk URLs after upgrade; hard reload fixes it
**Describe the bug**
After upgrading backlog.md, the browser UI stopped loading in my normal Chrome profile.

  The failing request looked like:

  GET http://localhost:6420/chunk-t1h94j9b.js -> 404

  At the same time, the running server was serving a different current asset set. For example:

  - GET / returned HTML referencing ./chunk-vbg3wk2g.js
  - GET /chunk-vbg3wk2g.js returned 200
  - GET /chunk-t1h94j9b.js returned 404

  So the browser appeared to be using a stale cached app shell or stale chunk references after upgrade.

**To Reproduce**
 1. Open backlog browser in Chrome on one version
  2. Upgrade backlog.md
  3. Open backlog browser again in the same profile
  4. Observe stale chunk request(s) with 404
  5. Use incognito or hard reload and observe that the app then works

**Expected behavior**
Upgrading backlog.md should not leave an existing browser profile stuck requesting old chunk filenames. If frontend assets change, the app should recover without requiring a manual hard reload.

**Environment**
 - backlog.md 1.44.0
  - global npm install under NVM
  - Linux
  - Google Chrome
  - local URL: http://localhost:6420
 - node v20.20.0

**Additional context**

MrLesk:
## Summary

Fixes #611.

This prevents the browser UI from staying on stale frontend asset references after a Backlog.md upgrade.

- Mark Bun's compiled SPA HTML bundle entry as `no-store` while leaving hashed JS/CSS asset behavior intact.
- Keep the existing Bun HTML import serving model instead of replacing the static serving stack.
- Add an inline `index.html` shell refresh guard that fetches the current page with `cache: "no-store"`, compares stylesheet/module asset references, and reloads once with `__backlog_reload` when the loaded shell is stale.
- Add focused regression coverage for the shipped shell guard, HTML bundle header mutation, and unchanged `/assets/*` cache behavior.

## Root Cause

The SPA HTML routes were served as Bun route values, so exact route responses bypassed the fallback `fetch` handler where no-cache headers were being set. In production, Bun HTML imports become a manifest with per-file headers, so stale browser profiles could retain an old app shell that referenced obsolete hashed chunks.

## Validation

- `bun test src/test/server-cache.test.ts src/test/server-assets.test.ts`
- `bunx tsc --noEmit`
- `bun run check .`
- `bun run build`
- Manual compiled binary check: `/` returns no-store headers and includes the shell-refresh guard; obsolete chunk URLs remain normal 404 fallback responses.
- `bun test`

### DIFF
--- a/backlog/tasks/back-395 - Web-prevent-stale-SPA-loads-by-disabling-index.html-caching.md
+++ b/backlog/tasks/back-395 - Web-prevent-stale-SPA-loads-by-disabling-index.html-caching.md
@@ -1,15 +1,21 @@
 ---
 id: BACK-395
 title: 'Web: prevent stale SPA loads by disabling index.html caching'
-status: To Do
+status: Done
 assignee:
   - '@codex'
 created_date: '2026-02-22 12:32'
+updated_date: '2026-04-26 08:36'
 labels: []
 dependencies: []
 references:
   - src/server/index.ts
   - src/web/index.html
+  - 'https://github.com/MrLesk/Backlog.md/issues/611'
+modified_files:
+  - src/server/index.ts
+  - src/test/server-cache.test.ts
+  - src/test/server-assets.test.ts
 priority: high
 ---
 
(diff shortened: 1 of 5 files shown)


MrLesk:
@codex please review this PR.

PR #625, opened 2026-05-02
BACK-458 - Fix promoted draft default status

### LINKED ISSUE #624: [Bug]: Draft tasks not working
**Describe the bug**
When a draft is promoted to a task, its state is still "Draft", even if the statuses do not include a "Draft" state. To my understanding, when promoting a draft it should set to the value of "default_status". Otherwise, the new task is not visible at all in the Kanban board.

Of course, its possible to add a "Draft" status as a workaround to see promoted drafts in the kanban board, but then the whole Drafts feature is not needed any more, because draft tasks are visible in the kanban board.

Or I have not understood the idea behind the fraft section at all...

**To Reproduce**
1. Install backlog.md with default states
2. Create draft. promote it to a task.
3. Promoted task not visible in kanban board

**Expected behavior**
I expect that the promoted draft is set to the value of the default_status.

**Environment**
- OS: Ubuntu Linux
- Node version: 1.44.0

**Additional context**
-

SAY-5:
Closes #624.

Promoting a draft now resets the draft-only `Draft` status to the configured `defaultStatus`, falling back to the built-in `To Do` status when no default is configured. This keeps promoted drafts visible as regular tasks on the kanban board.

The PR now includes the Backlog task metadata for `BACK-458` and focused regression coverage for both fallback and configured-default behavior.

### Validation
- `bun test src/test/filesystem.test.ts`
- `bunx tsc --noEmit`
- `bun run check .`
- `git diff --check`

### DIFF
--- a/backlog/tasks/back-458 - Fix-promoted-draft-default-status.md
+++ b/backlog/tasks/back-458 - Fix-promoted-draft-default-status.md
@@ -1,0 +1,71 @@
+---
+id: BACK-458
+title: Fix promoted draft default status
+status: Done
+assignee:
+  - '@codex'
+created_date: '2026-05-02 15:28'
+updated_date: '2026-05-02 15:30'
+labels:
+  - bug
+dependencies: []
+references:
+  - 'https://github.com/MrLesk/Backlog.md/issues/624'
+  - 'https://github.com/MrLesk/Backlog.md/pull/625'
+modified_files:
+  - src/file-system/operations.ts
+  - src/test/filesystem.test.ts
+priority: medium
+---
+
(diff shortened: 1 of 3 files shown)


MrLesk:
@codex please review this PR. If it is ready to merge, please add a 👍 reaction to the main PR description.

MrLesk:
@codex review

I addressed the P1 feedback: promoted draft-only tasks still use defaultStatus/FALLBACK_STATUS, but demoted tasks with an existing non-draft status now keep that status when promoted again. Added regression coverage for the demote/promote round trip.

PR #627, opened 2026-05-03
BACK-450 - Add CLI document update command

MrLesk:
## Summary
- add `backlog doc update <docId>` for content/title/type/tags/path updates
- preserve existing document body and metadata when options are omitted
- update CLI and agent guidance for the new document update path

## Task
- BACK-450 - Add CLI document update command

## Validation
- bun test src/test/cli.test.ts
- bunx tsc --noEmit
- bun run check .


MrLesk:
@codex review

PR #631, opened 2026-05-03
BACK-464 - Use shared multi-label filter dropdown on board

MrLesk:
## Summary
- Extract the All Tasks labels dropdown into a shared `LabelFilterDropdown` component.
- Reuse the shared multi-select labels dropdown on the Kanban board instead of the native single-select label filter.
- Persist Board labels as repeated `label` URL params and filter by any selected label while preserving assignee/priority behavior.
- Pass configured `availableLabels` into Board so the dropdown has the same label source family as All Tasks.

## Tests
- `bun test src/test/web-board-filters.test.tsx src/test/web-task-list-labels-menu.test.tsx`
- `bunx tsc --noEmit`
- `bun run check .`

### DIFF
--- a/backlog/tasks/back-464 - Use-shared-multi-label-filter-dropdown-on-board.md
+++ b/backlog/tasks/back-464 - Use-shared-multi-label-filter-dropdown-on-board.md
@@ -1,11 +1,11 @@
 ---
 id: BACK-464
 title: Use shared multi-label filter dropdown on board
-status: In Progress
+status: Done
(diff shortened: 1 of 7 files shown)


MrLesk:
Alex's Agent: @codex review

MrLesk:
Alex's Agent: Addressed the Codex label-normalization feedback in e1ac283 so configured label casing matches task labels case-insensitively. @codex review

PR #638, opened 2026-05-07
BACK-508 - Fix cli-commit-behaviour tests using git rev-list --count HEAD

lenucksi:
Basic problem here: If you use a tool that also writes to separate refs per commit, this will break a bunch of tests.

Namely these here:
```
✗ CLI Auto-Commit Behavior with autoCommit: true > should commit when creating a task if autoCommit is true [608.01ms]
✗ CLI Auto-Commit Behavior with autoCommit: true > should commit when creating a document if autoCommit is true [574.01ms]
✗ CLI Auto-Commit Behavior with autoCommit: true > should commit when creating a decision if autoCommit is true [581.01ms]

```
This ensures they're not breaking in such contexts.

## Summary

- `getCommitCountInTest` in `src/test/cli-commit-behaviour.test.ts` used `git rev-list --all --count`, which includes commits in `refs/notes/ai` — a ref written asynchronously by a global `git-ai` daemon configured via `trace2.eventTarget` in `~/.gitconfig`.
- Because the daemon writes note commits after a short delay, the count taken before the CLI command (`initialCommitCount`) and after it (`finalCommitCount`) could differ by 1 even when no real commit was made, causing all 6 tests to fail non-deterministically.
- The fix is a one-character change: `--all` → `HEAD`. `git rev-list --count HEAD` counts only commits reachable from the current branch tip, which is the correct metric for "did this CLI command produce a commit on this branch?". This matches the approach already used by the sibling tests in `cli.test.ts`.
- No implementation changes are needed; `shouldAutoCommit` and all three CLI handlers (`task create`, `doc create`, `decision create`) already read `autoCommit` from config correctly.

## Test plan

- [x] `bun test src/test/cli-commit-behaviour.test.ts` — 6/6 pass (was 0/6)
- [x] `bun test` — 1256 pass, 0 fail, no regressions
- [x] `bunx tsc --noEmit` — clean
- [x] `bun run check .` — clean


MrLesk (changes requested):
Alex's Agent: Thanks, the `git rev-list --count HEAD` change itself looks reasonable for this test. Counting only commits reachable from the current branch is the right signal for these auto-commit assertions, and it avoids unrelated refs such as notes.

I do not think this PR is ready to merge as-is though, because the branch includes unrelated terminal-status configuration changes in addition to the `BACK-471` test fix.

Could you please split or rebase this PR so it only contains the `src/test/cli-commit-behaviour.test.ts` change?

Specific blockers:

1. The PR title/body describe a one-line test fix and say no implementation changes are needed, but the branch also changes `src/file-system/operations.ts`, `src/types/index.ts`, `src/utils/terminal-status.ts`, and `src/test/terminal-status.test.ts` for `terminalStatuses`.

2. The included `terminalStatuses` config work is incomplete. It parses `terminal_statuses`, but `serializeConfig` does not write it back, so a later config save can silently drop that setting.

3. The new helper accepts a `terminalStatuses` override, but the actual runtime call sites still pass only `statuses`: CLI cleanup, core cleanup selection, the Web board cleanup affordance, and the All Tasks terminal-status filtering. So the tests cover the helper directly, but not behavior users would actually see.

4. I do not see a matching Backlog task for `BACK-471` in this branch. If this PR is kept, it should include the task details according to the project workflow.

I would be happy with the `HEAD` test-count change as a narrow PR once the unrelated terminal-status changes are removed or moved into a separate, complete task/PR.

lenucksi:
Branch has been force-pushed and cleaned up. The PR now contains exactly one commit with two files:

- `src/test/cli-commit-behaviour.test.ts` — the single-line fix (`--all` → `HEAD`)
- `backlog/tasks/back-471 - Fix-CLI-auto-commit-for-doc-decision-task-create-commands.md` — task record (status: In Review)

All unrelated terminal-status changes from BACK-466/467 have been removed. `bun test src/test/cli-commit-behaviour.test.ts` passes (6/6).

MrLesk:
Thanks for the cleanup @lenucksi 
In the next days I will try to go over all the PRs

PR #651, opened 2026-05-13
BACK-473 - handle port congestion for backlog browser

lenucksi:
## Summary
- Add `isPortAvailable()` and `findNextAvailablePort()` helpers to `src/server/index.ts`
- Pre-check port availability in CLI browser command before calling `server.start()`
- If port is taken: suggest next free port, prompt user interactively (readline), start there if accepted
- Simplify EADDRINUSE catch block in `BacklogServer.start()` (UX handled upstream in CLI)
- Unit tests for port helpers (5 cases, including 2 error/edge cases)

## Test plan
- [ ] `bun test src/test/server-port.test.ts` passes (5/5)
- [ ] `bun test` — no new failures (1244 pass, 2 skip, 4 pre-existing fail)
- [ ] `bunx tsc --noEmit` passes
- [ ] `bun run check .` passes

### DIFF
--- a/backlog/tasks/back-473 - handle-port-congestion-for-backlog-browser.md
+++ b/backlog/tasks/back-473 - handle-port-congestion-for-backlog-browser.md
@@ -3,12 +3,18 @@ id: BACK-473
 title: handle port congestion for backlog browser
 status: Done
 assignee:
-  - '@claude'
+  - '@codex'
 created_date: '2026-05-08 14:29'
-updated_date: '2026-05-13 18:52'
+updated_date: '2026-07-08 19:55'
 labels:
   - webui
 dependencies: []
+modified_files:
+  - src/server/index.ts
+  - src/cli.ts
+  - src/test/server-port.test.ts
+  - src/test/cli-browser-port.test.ts
+  - src/test/test-utils.ts
 ordinal: 166000
 ---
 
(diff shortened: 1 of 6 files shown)


MrLesk (changes requested):
Alex's Agent: Thanks for the patch. I’m requesting changes before this is ready to merge.

Focused validation on a merge with latest `origin/main` passed:
- `bun test src/test/server-port.test.ts`
- `bunx tsc --noEmit`
- `bun run check .`

Blockers:
- `findNextAvailablePort(port + 1)` has no upper bound. If the selected port is `65535` and occupied, or the scan reaches `65536`, `isPortAvailable()` returns `false` forever and the CLI hangs.
- `backlog browser` now prompts when the configured port is occupied even in non-TTY/headless/piped contexts. This can hang automation unless users already know to pass the new `--non-interactive` flag.

Please bound the port scan/fail cleanly, and make non-TTY/headless use non-interactive by default or keep a clean exit path for those cases. Add coverage for those paths.

For future PRs, please link the relevant GitHub issue first where applicable, then include Backlog task IDs as secondary context.

MrLesk:
Alex's Agent: I re-reviewed this against current main. This is a real narrow browser-port bug, but the existing requested changes are still valid. Please bound the port scan so values beyond `65535` cannot loop forever, make non-interactive/headless behavior explicit instead of prompting by default, and avoid fixed high ports in the tests because they can collide on shared CI/dev machines. The PR currently has no GitHub check evidence.

MrLesk:
@codex please review

PR #687, opened 2026-06-14
BACK-508 - Resolve MCP project root from client workspace roots

### LINKED ISSUE #558: [Bug]: MCP server writes to main repo instead of git worktree
## Description

When using Backlog.md MCP server from within a git worktree, task files are created in the main repository's `backlog/tasks/` directory instead of the current worktree's directory.

## Steps to Reproduce

1. Set up a project with Backlog.md and a protected `main` branch
2. Create a git worktree: `git worktree add .worktrees/feature-branch -b feature-branch main`
3. Open Claude Code (or any MCP client) with CWD set to the worktree
4. Create a task via the MCP `task_create` tool

**Expected:** Task file created in `.worktrees/feature-branch/backlog/tasks/`
**Actual:** Task file created in the main repo's `backlog/tasks/`

## Impact

Since `main` is protected, the task files created by MCP can't be committed. The workaround is to manually copy files from the main repo to the worktree, then delete the originals — which defeats the purpose of MCP automation.

## Suggested Fix

Resolve the project root using `process.cwd()` or `git rev-parse --show-toplevel` at the time of each file operation, rather than at server startup. In a worktree, `--show-toplevel` correctly returns the worktree path.

## Environment

- Backlog.md: latest (installed via homebrew)
- MCP config: `backlog mcp start` (stdio, user scope)
- Git worktrees via `git worktree add`
- macOS

ycaptain:
## Summary
The MCP server chose its project root once at startup and only looked at the client's MCP roots when it had no project of its own (fallback mode). A server shared across projects, or one started in the main checkout while the active workspace is a git worktree, therefore kept writing tasks to whatever directory it launched in (#558).

This makes an already-initialized server follow the client workspace as well, reusing the request-scoped roots discovery that #608 added for fallback mode. On its first request the server asks the client for its roots and switches to the matching project, then re-checks on every `roots/list_changed`. If the workspace has no backlog of its own, the server stays on its launch directory instead of dropping to "init required". Passing `--cwd` or `BACKLOG_CWD` pins the root and turns roots discovery off, which is what a deliberately shared server wants.

## Behavior change
An initialized, unpinned server now sends one `roots/list` request on its first handled request, where before it sent none. If the client's root matches the launch directory, nothing is re-registered. Clients that do not advertise the roots capability keep the old `process.cwd()` behavior. A pinned server sends no roots requests at all.

## Implementation notes
* `src/commands/mcp.ts` marks the root as pinned when it comes from `--cwd` or `BACKLOG_CWD`, and unpinned when it is only `process.cwd()`.
* `src/mcp/server.ts` adds a `startupHasProject` flag so an initialized server returns to its own project when the client offers nothing usable, while a fallback server still reverts to init-required.
* `upgradeToProject`, `downgradeToFallback` and `resolveFromRoots` are unchanged, as are the root checks from #570: each root is checked on its own, no walking up the tree, one resolution in flight at a time.
* README now documents workspace following and the `--cwd` / `BACKLOG_CWD` pin.

## Related Tasks
Closes #558. Task `backlog/tasks/back-558 - Resolve-MCP-project-root-from-client-workspace-roots.md` (BACK-558), acceptance criteria and DoD checked.

## Validation
* `bun test src/test/mcp-workspace-root.test.ts src/test/mcp-roots-discovery.test.ts` (15 pass): following the client workspace from an unrelated launch directory, a pin that sends zero roots requests, a monorepo package kept, the launch project kept when the workspace has no backlog, the return to the launch project after the workspace loses its backlog, and the worktree case from #558.
* `bunx tsc --noEmit`
* `bun run check .`


MrLesk:
Thanks @ycaptain 
This is a really nice fix

PR #701, opened 2026-06-27
feat: authenticated multi-user hosting — per-user commit attribution (web) + Streamable HTTP MCP transport

tuxo83:
# feat: authenticated multi-user hosting — per-user commit attribution (web) + Streamable HTTP MCP transport

## Summary
Two complementary changes that let Backlog.md be hosted for a team behind a single OIDC
provider, with every commit attributed to the authenticated end-user:

1. **Web — commit author from auth-proxy headers.** When the web server runs behind an
   authenticating reverse proxy (e.g. oauth2-proxy), the commit **author** is derived
   from the forwarded identity headers; the **committer** stays the server identity.
   Opt-in, header names configurable (oauth2-proxy defaults).
2. **MCP — Streamable HTTP transport.** `backlog mcp start --http` serves the MCP server
   over HTTP as a single long-lived process (no per-client stdio process), behind the
   same proxy, reusing the **same** per-request identity → commit-author mechanism.

This PR contains two reviewable commits (web first, then the MCP transport that builds on
it). They are submitted together because the MCP transport reuses the web commit's new
`commit-context.ts` and the shared `authorFromProxyHeaders()` helper, so they share config
plumbing and cannot be opened as independent cross-fork PRs.

## Commit 1 — web commits attributed to the auth-proxy user
- New opt-in config `commit_author_from_proxy_headers` (default off) with configurable
  header names `proxy_author_email_header` / `proxy_author_name_header`
  (defaults `x-forwarded-email` / `x-forwarded-preferred-username`).
- A small `AsyncLocalStorage`-backed `commit-context.ts` binds a per-request git author;
  the commit builders append `--author` when one is set. Concurrency-safe, no signature
  threading through `Core`.
- The web server wraps mutating routes so each commit is authored by the proxied user.

## Commit 2 — Streamable HTTP transport for MCP
- `backlog mcp start --http [--port 6421] [--host 127.0.0.1]`; stdio stays the default.
- Uses the SDK's `WebStandardStreamableHTTPServerTransport` (no new dependency), stateless
  (fresh transport + watchers-disabled server per request) → no per-client process spawned.
- Per-request author resolution is shared with the web server via `authorFromProxyHeaders()`
  (extracted from the web server, which now reuses it); request handling is wrapped in
  `runWithCommitAuthor()`.
- `AUTHENTICATED-HOSTING.md`: end-to-end guide (Keycloak + oauth2-proxy, web + MCP, and the
  two MCP client auth options — Bearer token for agents, interactive OAuth for humans).

## Behavior & safety
- Both behaviours are **opt-in** and default off — no change for existing users.
- The git **committer** always stays the server identity; only the **author** is overridden,
  so history remains auditable.
- The config field is intentionally **not** added to `needsMigration` (no forced config
  rewrite on existing projects).

## Testing
- `bunx tsc --noEmit` → clean. `npx biome check .` (changed files) → clean.
- `bun test`:
  - `commit-author.test.ts` — author from the request context, committer preserved, fallback
    to git identity.
  - `mcp-http-transport.test.ts` — end-to-end via the MCP SDK client over HTTP: a tool call
    with forwarded identity headers produces a commit authored by the proxied user (committer
    preserved), and falls back to the git identity without headers.
  - `config-commands.test.ts` — config serialize/parse roundtrip.

### DIFF
--- a/AUTHENTICATED-HOSTING.md
+++ b/AUTHENTICATED-HOSTING.md
@@ -1,0 +1,192 @@
+# Authenticated multi-user hosting
+
+This guide shows how to host Backlog.md as a shared service for a team — the **web UI**
+and the **MCP server** — behind a single OpenID Connect (OIDC) provider, so that every
+commit is attributed to the authenticated end-user.
+
+It uses [Keycloak](https://www.keycloak.org/) as the example identity provider and
+[oauth2-proxy](https://oauth2-proxy.github.io/oauth2-proxy/) as the authenticating
+reverse proxy, but any OIDC provider / proxy that forwards identity headers works.
+
+> Backlog.md never talks to the identity provider itself. It trusts the identity that
+> the proxy forwards as request headers, and turns it into the **git commit author**.
+> The git **committer** always stays the server's own identity.
+
+## Architecture
+
+```
+                       ┌──────────────┐
+                       │   Keycloak   │  (OIDC provider)
+                       └──────┬───────┘
+                              │ OIDC
+                   ┌──────────┴───────────┐
+                   │     oauth2-proxy     │  forwards X-Forwarded-Email /
+                   └───┬──────────────┬───┘  X-Forwarded-Preferred-Username
+                       │              │
+              ┌────────┴───┐   ┌──────┴───────────────┐
+              │ backlog    │   │ backlog mcp start    │
+              │ browser    │   │ --http               │
+              │ (web UI)   │   │ (Streamable HTTP)    │
+              └────────┬───┘   └──────┬───────────────┘
+                       └───────┬──────┘
+                        ┌──────┴───────┐
+                        │  Backlog.md  │  auto_pull / auto_push
+                        │  git repo    │  + push deploy key
+                        └──────┬───────┘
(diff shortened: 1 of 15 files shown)


MrLesk:
Alex's Agent: Thanks for the detailed PR. I’m closing this because it does not fit the current project direction: we are keeping the product small, CLI/MCP-first, and git-authorized.

This PR adds authenticated multi-user hosting semantics, a web auth/proxy trust surface, per-user hosted commit attribution, a new Streamable HTTP MCP transport, and hosted setup docs/config around Keycloak/oauth2-proxy. Even though it is opt-in, that is a broad product and security surface we are not taking into the current repository scope.

I did not find a small independent bugfix in the diff that should be split out separately. If there is a narrower local CLI/MCP bugfix behind this work, please open that as a separate focused PR.

kenorb:
I've written relevant [`backlog-browser-host.sh`](https://gist.github.com/kenorb/5d1c242ec9205d1bb7a5c8fc3f487aff) script which exposes localhost board on the host while running in the devcontainer/Docker containers which can be helpful.

Issue #684, opened 2026-06-11
`updated_date` gets bumped when only the ordinal changes


BenNewman100:
Hello, I ran into this while migrating a project onto Backlog.md (really liking the tool, btw). Might be intended, but figured I'd flag it.

Changing a task's `ordinal` (just its sort position) also rewrites `updated_date`. That's a bit annoying, as reordering isn't really "updating" the task content. I was planning on using the `updated_date` field to detect staleness.

**Repro (v1.45.2):**
```sh
mkdir demo && cd demo && git init
backlog init demo --agent-instructions none
backlog task create "Alpha"
backlog task edit task-1 --ordinal 500   # change ONLY the ordinal
```
Before / after:
```diff
  created_date: '2026-06-10 19:16'
+ updated_date: '2026-06-10 19:16'
- ordinal: 1000
+ ordinal: 500
```
A fresh task has no `updated_date`, and the ordinal-only edit adds one. Board drag-reorder does the same.

**Where it bit me:** hitting **Sort** in the board UI reassigns ordinals across the whole board (totally expected), but the same pass also re-stamps `updated_date` on every task. In my case that overwrote ~95 distinct real dates with one identical timestamp. The ordinal changes are fine, losing the dates is the annoying part.

Would be nice if an ordinal-only change left `updated_date` alone, but curious whether that's intentional.

kuwork:
If the sequence number is changed, the updated_date should be updated as well. To solve this issue, I added support for temporary sorting.

<img width="321" height="324" alt="Image" src="https://github.com/user-attachments/assets/bd0833a2-6dab-43d7-800c-5c815428a045" />

I originally added those extra time fields for Gantt chart tracking, but as it turns out, they also solved the problem you brought up.

<img width="274" height="363" alt="Image" src="https://github.com/user-attachments/assets/c52a2f98-fdde-46bf-8926-d6b73eee29fe" />

BenNewman100:
I mean, to be fair, I still think that the ordinality number changing shouldn't cause "updated_date" to be changed. Or there should be another "last_modified" date that only applies to the task content. (not ordinality or priority) 

Ordinality and Prioity things that matter in the context of the whole project, they really are a concern of the task itself.

MrLesk:
[...]

PR #707, opened 2026-07-01
Issue #684 investigation prototype (closed pending owner decision)

MrLesk:
Alex's Agent:

Correction: this PR is not an accepted fix. It was opened during investigation of #684, but the product semantic needs an owner decision before any implementation should advance.

Prototype summary:
- Preserves existing updated_date, or leaves it absent, when task edit or board reorder changes only ordinal.
- Keeps updated_date refreshes for mixed edits that also change content, status, or milestone.
- Adds no new temporal fields.

Validation from the prototype branch:
- bun test src/test/core.test.ts src/test/reorder-utils.test.ts src/test/cli-plain-create-edit.test.ts
- bunx tsc --noEmit
- bun run check .
- bun test src/test/cli-priority-filtering.test.ts
- bun test src/test/cli.test.ts --test-name-pattern "should filter tasks by status case-insensitively"

This PR should remain closed/not for merge unless the owner explicitly accepts this semantic.

### DIFF
--- a/backlog/tasks/back-511 - Investigate-issue-684-updated_date-bump-on-ordinal-only-edits.md
+++ b/backlog/tasks/back-511 - Investigate-issue-684-updated_date-bump-on-ordinal-only-edits.md
@@ -1,0 +1,62 @@
+---
+id: BACK-511
+title: 'Investigate issue #684 updated_date bump on ordinal-only edits'
+status: Done
+assignee:
+  - '@codex'
+created_date: '2026-07-01 17:57'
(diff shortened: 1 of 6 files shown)


MrLesk:
Alex's Agent: Correction: this PR was opened before owner decision on the ambiguous updated_date/ordinal metadata semantic. I am closing it so it is not treated as an accepted fix. The prototype can be revisited if the owner accepts preserving updated_date for ordinal-only changes.

PR #715, opened 2026-07-02
BACK-518 - Preserve updated_date for ordinal-only task reorders

### LINKED ISSUE #684: `updated_date` gets bumped when only the ordinal changes

MrLesk:
Alex's Agent: This PR keeps `updated_date` focused on meaningful task edits by preserving the existing value when only `ordinal` changes.

Fixes #684.

Summary:
- Centralizes the timestamp decision in `Core.updateTask` by comparing persisted task fields while ignoring `ordinal` and `updatedDate`.
- Preserves existing `updated_date`, including absence, for direct ordinal edits, bulk ordinal updates, and same-column reorder flows.
- Keeps normal timestamp updates when ordinal changes are saved with content or metadata edits.

Validation:
- `bun test src/test/reorder-utils.test.ts`
- `bunx tsc --noEmit`
- `bun run check .`
- `bun test` (1379 pass, 2 skip)

### DIFF
--- a/backlog/tasks/back-518 - Preserve-updated_date-for-ordinal-only-task-reorders.md
+++ b/backlog/tasks/back-518 - Preserve-updated_date-for-ordinal-only-task-reorders.md
@@ -1,0 +1,51 @@
+---
+id: BACK-518
+title: Preserve updated_date for ordinal-only task reorders
+status: Done
+assignee:
+  - '@codex'
+created_date: '2026-07-02 20:39'
+updated_date: '2026-07-02 20:46'
+labels: []
+dependencies: []
+references:
+  - 'https://github.com/MrLesk/Backlog.md/issues/684'
+priority: high
+ordinal: 113000
+---
+
+## Acceptance Criteria
+<!-- AC:BEGIN -->
+- [x] #1 Editing only a task ordinal preserves the existing updated_date value
+- [x] #2 Editing only a task ordinal does not add updated_date when it was absent
+- [x] #3 Saving ordinal changes together with any non-order task field updates updated_date normally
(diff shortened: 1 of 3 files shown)


MrLesk:
Alex's Agent: Reviewed PR #715. The implementation matches the owner decision: ordinal-only saves preserve the existing `updated_date`, including absence, while ordinal plus any meaningful task/content change still updates it normally. Direct edits, bulk updates, drag reorder, and column sort all route through the centralized `Core.updateTask` decision, and focused regression coverage is in place. CI is green.

Issue #711, opened 2026-07-01
[Feature]: Collision-free task ID mode — sequential IDs inherently collide across git clones


gaugau3000:
**Problem**

Sequential ID allocation (`max + 1`) cannot be made safe in Backlog.md's own target topology: multiple writers (humans on separate machines + AI agents on a server) sharing one backlog through a git remote. Each clone computes `max + 1` against the state it can see, so two writers creating tasks between one another's pushes deterministically mint the same ID.

The existing mitigations each solve a narrower problem:
- `withCreateLock()` (#565, and #620 for milestones) — a `proper-lockfile` mutex, so it only serializes writers **on the same filesystem**. It cannot span git remotes.
- `remote_operations: true` — fetching before allocating narrows the race window but doesn't close it: two writers allocating between each other's fetch and push still collide, and offline creation collides by design.

We hit this in production twice (team of humans + a server-side agent creating tasks concurrently): same `MEM-N` allocated to two different tasks, surfacing as pre-commit/push conflicts and, worse, ambiguous references from task dependencies.

**Proposal**

An opt-in config, e.g. `idGeneration: random`, that mints IDs as `<prefix>-<8 hex>` (`crypto.randomBytes(4)`). Random IDs need zero coordination: they work identically across machines, offline, and for concurrent agents — which fits the git-native design better than any locking can.

**Why it's low-risk — verified empirically on 1.x**

We've been running exactly this scheme in production (via a wrapper that authors the task file directly) alongside ~600 legacy numeric IDs:
- Non-numeric IDs already work everywhere: `list`, `view`, `edit`, `board`, `search`, `archive`, dependencies, subtask `-p`.
- The sequential allocator ignores non-numeric IDs when computing `max + 1`, so random and legacy numeric IDs coexist with no migration.
- One caveat to document: IDs are uppercased on output and looked up case-insensitively, so consumers parsing output should accept uppercase hex.

The only thing missing is native generation — today the file must be authored directly, because `task create` offers no `--id` and create-then-rename leaves a raceable transient numeric name.

Happy to contribute a PR if you're open to the direction.

MrLesk:
Alex's Agent: Thanks for the detailed report. I agree this is distinct from #710: that PR fixed create-time allocation across same-repository git worktrees, while this issue is about independent git clones and offline/concurrent writers. A filesystem/common-git-dir lock cannot coordinate those clones, and `remote_operations` can reduce but not eliminate the race.

[...]

PR #749, opened 2026-07-09
BACK-516 - Human-first duplicate task ID recovery

MrLesk:
Alex's Agent: This PR adds a human-first, CLI-canonical workflow for diagnosing and repairing duplicate task IDs without assuming an AI agent is present.

## Summary

- Detects active/completed, zero-padding-equivalent, huge numeric, dotted, and exact legacy ID collisions while preserving intentional archive reuse.
- Fails closed for ambiguous ID reads and mutations across CLI, core, server, browser, TUI, and the MCP adapter.
- Adds `backlog doctor` with exact paths, deterministic preview, explicit confirmation, safe `--fix --yes` automation, manual reference-review reporting, and diagnostic-only cross-branch findings.
- Adds the shared browser recovery flow with a second confirmation, route/history preservation, and modal focus containment.
- Preserves current main behavior for task-type API/configuration, latest-request-wins data loading, URL query state, single-flight type edits, and the two-row board controls.

## Ownership-safe rollback

Rollback records each staged file using device, inode, and byte-level SHA-256 identity. It quarantines installed targets before checking ownership, deletes only transaction-owned content, and restores backups with no-replace hard links. Deterministic race coverage proves that a concurrently recreated source or externally changed installed target is preserved, with actionable recovery paths and retained backups when reconciliation is required.

## Verification

- Pushed head: `ac7e295c2b73a4d23eee634fbafcb1c8e2e7ee17`
- Focused composed matrix: 248 passed, 0 failed, 1,429 assertions.
- Full suite: 1,643 passed, 2 expected interactive skips, 0 failed, 6,700 assertions across 189 files.
- `bunx tsc --noEmit`
- `bun run check .`: 324 files, no fixes required.
- `bun run build`
- `git diff --check`
- Compiled CLI smoke: deterministic TASK-2/TASK-3 preview and repair, two ambiguous references retained for human review, post-repair doctor clean.
- Compiled Chrome QA: desktop-primary at 1920px and best-effort 390px; warning, overlay, confirmation, focus, task-type filter, and two-row controls verified; no horizontal document/dialog overflow; zero console warnings or errors; successful repair refresh and no transaction artifacts.

## Scope

Related to #711. This PR intentionally stops at human-first collision diagnosis and recovery; collision-prevention policy and random identifiers are not included. #711 remains open.

### DIFF
--- a/backlog/tasks/back-257 - Deep-link-URLs-for-tasks-in-board-and-list-views.md
+++ b/backlog/tasks/back-257 - Deep-link-URLs-for-tasks-in-board-and-list-views.md
@@ -1,15 +1,26 @@
 ---
 id: BACK-257
 title: Deep link URLs for tasks in board and list views
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@pr755-takeover'
 created_date: '2025-09-06 22:11'
+updated_date: '2026-07-10 13:43'
 labels: []
 dependencies: []
+references:
+  - 'https://github.com/MrLesk/Backlog.md/issues/754'
+  - 'https://github.com/MrLesk/Backlog.md/pull/755'
+modified_files:
+  - src/core/content-store.ts
+  - src/server/index.ts
+  - src/test/content-store.test.ts
+  - src/test/server-tasks-spa-fallback.test.ts
 ---
 
 ## Description
 
+<!-- SECTION:DESCRIPTION:BEGIN -->
 Background
 
 The web app supports deep links for Documentation and Decisions using SEO‑friendly routes like `/documentation/:id/:title` and `/decisions/:id/:title` that load the item and update the URL accordingly (see `App.tsx`, `DocumentationDetail`, `DecisionDetail`, and `sanitizeUrlTitle`).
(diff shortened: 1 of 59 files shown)


MrLesk (commented):
Alex's Agent: CHANGES_REQUESTED for exact head `e0e7c95b11e3cf682dfbca6346152b87432eae20`.

Release blockers:

1. **[P1] Make repair installation no-clobber.** `applyDuplicateTaskIdRepair` checks each destination, then later installs with `rename(stagedPath, targetPath)`. A human or other process can create that destination in the gap; POSIX rename replaces it and the repair reports success. The Backlog create lock cannot serialize external writers. Use an atomic no-replace install and add a deterministic race regression.

2. **[P1] Preflight `backlog board export`.** This path skips duplicate diagnosis and exports `core.loadTasks()` after its ID map has already collapsed exact duplicates. A two-file `TASK-1` fixture exited 0, printed `Total tasks: 1`, emitted no warning, and exported only Beta. Abort with the shared path-qualified diagnosis and a nonzero exit before writing.

3. **[P1] Keep keyboard focus inside the destructive Web flow.** Opening the repair dialog leaves focus on the background `Review repair` button; clicking `Continue` removes the focused element and drops focus to `BODY`. There is no modal focus trap or reliable restoration. Independently reproduced in JSDOM (`dialogContainsFocus: false` after both transitions), and the supplied preview screenshot visibly retains the focus ring behind the overlay. Move initial focus into the dialog, trap it, move focus to the final confirmation state, restore it on close, and cover the behavior.

4. **[P1] Restore Windows CI.** The new ambiguity test asserts a forward-slash substring while Windows correctly reports native backslash paths, so `unit-test-windows (1/3)` fails. Make the assertion/output portable. That shard also hit the known 30-second packaging timeout; rerun it after the deterministic failure is fixed and require the aggregate Windows check to pass.

Additional safety gaps:

5. **[P2] Preserve or reject duplicate subtask identity.** `TASK-1.1` plus `TASK-1.01` is currently previewed as a rename to top-level `TASK-2` while retaining `parent_task_id: TASK-1`. Use parent-aware allocation or block dotted-ID repair with a clear reason.

6. **[P2] Do not report an incomplete reference scan as “no references.”** Glob/read failures are converted to empty content, so repair can proceed and the CLI can claim that no textual references were found. Surface scan failures and block or explicitly mark the preview incomplete.

Positive evidence: the human-first wording, CLI-first doctor flow, active/completed and padding semantics, archive exclusion, cross-branch diagnostic-only behavior, shared server/Web fingerprint flow, and supplied desktop visual progression are directionally correct. Locally, 68 focused tests, TypeScript, Biome, build, and diff-check passed. The Browser plugin was unavailable in this review session, so page/console claims were limited to the supplied desktop captures plus DOM interaction testing; mobile was intentionally not reviewed.

Please rebase onto current `main`, address these findings, obtain green CI, and request a new exact-head review.

MrLesk:
Alex's Agent: Review remediation is complete at exact head `9dd0db5a2017ce036d3aecf1c90b287abe0bb952`, rebased directly on current `main` `f48225bd01247e78409abd3965ccc443e0793648`.

All six findings are addressed:

- Repair installation now atomically claims each destination without replacement; a late external destination returns `EEXIST`, preserves external content, and enters the existing rollback path. A deterministic no-clobber regression covers the race.
- `backlog board export` runs the shared path-qualified duplicate preflight before loading the collapsible view or touching the destination; the regression preserves sentinel output and exits nonzero.
- The shared desktop modal now starts on safe Cancel, traps Tab/Shift+Tab, focuses the final Repair action after staging, closes on Escape, restores the Review trigger on Escape/header close, and exposes visible focus treatment.
- Platform-sensitive ambiguity assertions are Windows-neutral. All three Windows shards, Windows compiled smoke, and the aggregate Windows check are green on this head.
- Dotted duplicates use validated parent-aware allocation (`TASK-1.01` → `TASK-1.2`); missing or mismatched parent identity blocks the entire group.
- Reference glob/read failures produce an explicit incomplete state, participate in the fingerprint, block repair, and never claim that no references exist in CLI or Web.

Validation on the exact base:

- Focused boundary matrix: 76 passed, 0 failed, 362 assertions.
- Full suite: 1,544 passed, 2 intentional skips, 0 failed; 5,377 assertions across 183 files.
- `bunx tsc --noEmit`, `bun run check .`, `bun run build`, and `git diff --check` pass.
- Compiled CLI board-export, doctor preview/fix/verify, and dotted-subtask smokes pass.
- Desktop Chrome keyboard/visual QA passed with zero console warnings/errors; mobile remains intentionally out of scope.
- GitHub CI and CodeQL are fully green, including the Windows aggregate.

Please perform a fresh exact-head review. This PR remains unmerged.

PR #770, opened 2026-07-11
BACK-536 - Add wrapNavigationToSearch config to opt out of the boundary search handoff

### LINKED ISSUE #768: [Feature]: Option to disable the boundary jump into the search input (vim-style j/k)
In the TUI task list and kanban board, pressing `j` on the last row or `k` on the first row (or `k` at the top of the focused detail pane) moves focus into the search input. The wrap is deliberate (back-399) and works well with arrow keys, but it surprises vim-style navigation: `j`/`k` are expected to stay inside the list, and `/` and `Ctrl+F` already focus search directly. There is currently no way to opt out.

Suggested solution: a boolean config key `wrapNavigationToSearch` (default `true`) that keeps the current behavior by default. When set to `false`, boundary navigation no longer hands focus to search in the task list, the detail pane, or the kanban board, and falls back to the circular wrap both views used before back-399 (introduced in task-248).

I have a small PR ready that implements this as a pure opt-out (default behavior unchanged, covered by tests). Happy to adjust the key name or scope to whatever fits the project.

kossakovsky:
## Summary

In the TUI, pressing `j` on the last row or `k` on the first row (or `k` at the top of the focused detail pane) hands focus to the search input. The wrap is deliberate (back-399), but it surprises vim-style navigation: `j`/`k` are expected to stay inside the list, and `/` and `Ctrl+F` already focus search directly.

This adds an opt-out config key that keeps today's behavior by default:

- `wrapNavigationToSearch` (boolean, default `true`), available through `backlog config get/set/list` and documented in ADVANCED-CONFIG.md
- when `false`, the handoff stops in the task list, the detail pane, and the kanban board (including empty columns)
- with the handoff disabled, boundary navigation falls back to the circular wrap both views used before back-399 (introduced in task-248)

## Related Issue or Task

Closes #768

Backlog task: BACK-536

## Task Checklist

## Testing

- `bunx tsc --noEmit` — passed
- `bun run check .` — passed (323 files)
- `bun test` — 1670 passed + 2 intentional skips / 0 failed (187 files)
- new coverage: wrap-disabled cases in `task-viewer-boundary-navigation.test.ts`, `resolveBoardBoundaryToSearch` cases in `board-ui.test.ts`, and a `config get/set/list` round-trip for the new key

### DIFF
--- a/ADVANCED-CONFIG.md
+++ b/ADVANCED-CONFIG.md
@@ -30,6 +30,7 @@ Running `backlog config` with no arguments launches the interactive advanced wiz
 | `defaultEditor`   | Editor for 'E' key | Platform default (nano/notepad) |
 | `defaultPort`     | Web UI port        | `6420`                        |
 | `autoOpenBrowser` | Open browser automatically | `true`            |
+| `wrapNavigationToSearch` | Move focus into TUI search when navigating past the first/last row | `true` |
 | `remoteOperations`| Enable remote git operations | `true`           |
 | `autoCommit`      | Automatically commit task changes | `false`       |
 | `bypassGitHooks`  | Skip git hooks when committing (uses --no-verify) | `false`       |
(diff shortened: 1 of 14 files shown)


MrLesk:
Alex's agent: Closing this one — the underlying request (#768) shipped on main via #866, but with a different shape than proposed here: rather than adding a config key, j/k now simply never enter the search input at boundaries while the arrow keys keep the handoff, so both vim and regular users get sensible behavior with no new configuration surface. Thanks for the clean implementation and for offering to rescope — the boundary-site map in your diff was genuinely useful when building the shipped version.

PR #789, opened 2026-07-15
Permanent task IDs: unify draft+task pool, stop recycling archived IDs

DBR77Red:
## Problem

Two behaviours let the same ID refer to different work over a project's life:

1. **Archiving frees an ID for reuse.** `getExistingIdsForType` scans `tasks/` and `completed/` but not `archive/`, and the code says so explicitly (`"Archived tasks are intentionally excluded - archived IDs can be reused"`). Archiving the highest-numbered task returns its number to the pool, so the next `task create` re-issues it. On one board this silently produced **three different `task-29`s**, which makes `backlog task 29` ambiguous and breaks every `Refs: task-29` in git history.
2. **Drafts are a separate ID space and get renumbered on promotion.** A draft is `draft-7` from its own counter; promoting it mints a brand-new `task-42`. A draft ID can never be cited durably, and the two counters drift independently.

## Change

An ID is issued **at most once for the life of a project**:

- Tasks and drafts share one monotonic pool (`getExistingIdsForType` folds in drafts + archived tasks + archived drafts).
- A draft is born with a task ID (`task-N` in `backlog/drafts/`) and keeps it through promote/demote.
- Drafts follow the **configured** task prefix, so a `prefixes.task: "JIRA"` project gets `JIRA-N` drafts out of the `JIRA-N` pool (they must share a prefix or the pool splits).
- Removes the `task-*.md → draft-*.md` prefix migration, which is incompatible with this layout.

Drafts stay off the kanban board (native Drafts section) and "complete" is unchanged — those were never the problem.

~13 semantic edits across `core/backlog.ts`, `file-system/operations.ts`, `utils/prefix-config.ts`, `utils/task-path.ts`. Rationale + migration notes for adopters in `PERMANENT-IDS.md`.

## Tests

- Rewrote the tests that encoded the old contract (archived-ID reuse, `draft-N` output/promotion).
- New `src/test/permanent-ids.test.ts` covers the invariants: unified pool, promote/demote keep the ID, archived tasks **and** drafts never recycle, custom prefix, and a draft-filename round trip (that last one caught a real `draft-task-2` filename bug during development).
- All green except two pre-existing git/worktree tests that also fail on `main` in my Windows environment.

## Notes for maintainers

- This changes default behaviour. Happy to gate it behind a config flag (e.g. `reuseArchivedIds` / unified draft IDs) if you'd rather preserve the current default — let me know your preference.
- Adopters with existing `draft-*.md` files perform a one-time `draft-N → task-M` rename; documented in `PERMANENT-IDS.md`. I did not auto-migrate on load (too surprising).
- Separately, I noticed `Core.demoteTask` moves a task into `drafts/` without rewriting its `status` (a demoted item can sit in drafts still marked `To Do`). Left untouched here — happy to file it separately.


MrLesk:
Alex's Agent: Thank you for taking the time to document the failure mode and build thorough tests around it.

Backlog.md has an explicit product contract here: archive is soft delete. Archiving ends the task ID's active identity and intentionally releases that ID for future use. This PR reverses that settled behavior by reserving archived IDs permanently.

The PR also combines that change with a separate draft identity and prefix migration. Existing `draft-*` files would become undiscoverable until users manually rename them. That needs its own scoped product decision and migration design.

Because the central behavior conflicts with accepted maintainer direction, and there is no accepted issue or Backlog task authorizing either change, we are closing this PR rather than asking you to rework it. The separate observation about demoted drafts retaining a non-Draft status can be evaluated independently if reported with a current-main reproduction.

Thank you again for the clear report and implementation effort.

PR #793, opened 2026-07-16
BACK-550 - Add append-plan option to task edit CLI

### LINKED ISSUE #792: task edit: --append-notes has no --append-plan counterpart
`task edit --append-notes` has no plan equivalent, so the implementation plan is the only section a CLI caller can replace but never extend.

The plumbing already exists below the CLI: `TaskEditArgs.planAppend`, `Core.appendImplementationPlan`, and the MCP tool's `planAppend` parameter. Only the CLI never wired a flag to it.

I have a patch that mirrors `--append-notes`, including its `hasEditFieldFlags` entry so that `--append-plan` on its own counts as an edit rather than falling through to the wizard on a TTY, with tests mirroring `append-implementation-notes.test.ts`. Want a PR?

jacobjove:
Closes #792.

## Summary

- add repeatable `backlog task edit --append-plan <text>` support through the existing shared plan-append pipeline
- apply appends after `--plan` replacement, preserving CLI order and blank-line separation while ignoring blank values
- document the option in task edit help and the canonical task-execution guide
- cover missing plans, multiline and whitespace handling, replacement plus append, and real PTY no-wizard behavior

## Verification

- `bun test` (1729 pass, 4 skip, 0 fail)
- `bunx tsc --noEmit`
- `bun run check .`
- `bun run build`


MrLesk:
@codex review

PR #796, opened 2026-07-16
BACK-563 - Scope autoCommit to exactly the files each write touches

### LINKED ISSUE #795: autoCommit stages the whole backlog dir and commits with no pathspec, sweeping in unrelated dirty state
With `autoCommit` on, `document_update` (and `create`) stages the whole backlog directory and then runs a bare `git commit -m <msg>` with no pathspec. Anything else already dirty in the working tree or index gets swept into that commit, under a message that describes only the document write.

`createDecision`, `updateTasksBulk`, `archiveTask`, and `completeTask` share the shape. Some stage narrowly but still finish with a bare, unscoped commit, which takes the working tree rather than what they staged. The five draft/task lifecycle sites (`promoteDraftWithUpdates`, `demoteTaskWithUpdates`, and friends) use `stageBacklogDirectory()` + bare `commitChanges()`, which also sweeps an untracked file that happens to live under `backlog/`.

We hit this with two agent sessions sharing a repo: one session's unreviewed draft landed inside another's "Add document" commit. A single user with a clean tree would never notice.

Two adjacent bugs in `commitFiles` that fall out of the same area:

- Without `--no-renames`, its internal staged-paths check collapses a staged delete+add pair (an archive move) into one rename entry naming only the destination, so the source path is silently dropped from the commit. This affects the existing `archiveMilestone`/`renameMilestone` callers.
- A requested path with no git history at all (created and archived in one operation while `autoCommit` was off) is not a valid pathspec and aborts the whole commit.

The fix we are running locally routes every one of these sites through `addFile`/`stageFileMove` + `commitFiles`, so each commit contains exactly the paths that write touched. It carries tests covering the sweep (a peer's dirty file must stay out of the commit) and both `commitFiles` cases. It applies cleanly to current `main` and the suite is green.

Want a PR?

jacobjove:
Closes #795.

Tracks BACK-563.

## Summary

- Scope document, decision, task, bulk-update, archive, completion, promotion, demotion, milestone, and agent-instruction auto-commits to the exact paths written or moved by each operation.
- Preserve unrelated staged, unstaged, and untracked work instead of sweeping it into Backlog.md commits or clearing shared index state.
- Reconcile the contributor patch with current `main` and its temporary-index, hook-aware, compare-and-swap HEAD ownership pipeline.
- Remove the unused pathless fallback from `commitTaskChange`; all production task callers now require an owned file path.
- Cover moved/deleted paths, paths with no Git history, non-ASCII filenames, and concurrent HEAD/index/worktree ownership.

The contributor commits are preserved, followed by a normal merge from current `main` and a focused integration commit.

## Verification

- `bunx tsc --noEmit`
- `bun run check .`
- `bun run build`
- focused scoped-ownership and concurrency matrix
- full suite: 1,851 passed, 5 skipped, 0 failed across 208 files


jacobjove:
Pushed a third commit, for one more site in this family.

`addAndCommitTaskFile()`, the auto-commit behind every task create and edit, has the shape #795 describes: it stages narrowly, then finishes with a pathspec-less commit. It also runs an unscoped `git reset HEAD` before staging, which does more than sweep. With a second session working in the repo it discards that session's staged work outright, and their next commit lands empty or short with nothing announcing the loss. We hit this one directly: a peer's `backlog task edit` cleared an index between our `git apply --cached` and our `git commit`.

It now stages and commits through `commitFiles()` like the other sites. `resetIndex()` and `commitStagedChanges()` have no callers left and are removed; the scoped `resetPaths()` sibling is untouched.

The reset came from BACK-163, whose acceptance criteria were that task edit "should only stage and commit the modified task file" and "should not commit any other staged or unstaged changes in the repository". Isolation was the goal and the reset was the mechanism; a scoped commit reaches the same goal without clearing an index it does not own.

The new test fails on the old code, where the peer's staged path comes back empty, and passes now. `bunx tsc --noEmit` clean and suite green.

PR #802, opened 2026-07-19
BACK-554 - Package Backlog.md with bun2nix v2

### LINKED ISSUE #801: [Bug]: `nix run` segfaults on x86_64-linux (in-flake `bun --compile`)
**Describe the bug**

`nix run github:MrLesk/Backlog.md` (and `nix build .#`) produces a binary that segfaults on startup for every command:

```
$ nix run github:MrLesk/Backlog.md -- --version
Segmentation fault (core dumped)
```

The crash happens before any shared library loads (`strace` shows only `execve → brk → SIGSEGV` at address `0x8`), so it is not the app code and not the CPU (reproduces on a modern AVX2 machine). The npm / GitHub-release binaries work fine — only the Nix build is affected.

Two independent causes, both in the in-flake `bun build --compile` path, reproduced with a one-line `console.log("hi")`:

1. The nixpkgs baseline-Bun overlay in `flake.nix` (added for #412) yields a standalone that segfaults. Non-baseline Bun compiles a working one; only the nixpkgs-processed baseline variant is affected.
2. `bun2nix`'s `mkBunDerivation` patchelfs the compiled Bun single-file executable to a Nix glibc, corrupting Bun's embedded trailer — this segfaults even with non-baseline Bun. Looks closely related to (possibly a dup of) #584.

| Bun as compiler | raw `--compile` | after mkBunDerivation patchelf |
|---|---|---|
| nixpkgs non-baseline | works | segfault |
| nixpkgs baseline (current flake) | segfault | segfault |
| raw oven-sh baseline | works | n/a |

**To Reproduce**
1. On x86_64-linux with Nix (flakes enabled), run `nix run github:MrLesk/Backlog.md -- --version`
2. See `Segmentation fault (core dumped)`

**Expected behavior**
`nix run` / `nix build .#` produces a working `backlog` binary, same as the npm and GitHub-release binaries.

**Environment**
- OS: Pop!_OS (Linux, kernel 6.9.3)
- Node version: n/a (Nix build; not run via Node)
- Nix: 2.34.8
- Bun: 1.3.14 (flake-pinned)
- glibc: 2.40
- Backlog.md: 1.48.0
- CPU: AMD Ryzen Threadripper 3960X (has AVX2)

**Additional context**
Related: #412 (origin of the baseline overlay), #584 (patchelf / hardcoded library paths). I could not find any way to run baseline Bun as an in-sandbox compiler (no `/lib64`; a loader wrapper breaks `--compile` with `BunSectionNotFound`; patchelf-ing baseline Bun makes Bun itself segfault), so a fix likely involves an architecture choice about how the flake is built (e.g. packaging the prebuilt release artifact vs. changing the in-flake compile) and how #412 old-CPU support is preserved. Flagging that rather than prescribing it; happy to share a working proof-of-concept.

MrLesk:
## Summary

- replace the legacy Nix path with the current bun2nix v2 fetchBunDeps and mkDerivation workflow
- package the CLI bundle and generated browser assets behind a Nix-managed Bun runtime
- support x86_64-linux, aarch64-linux, and aarch64-darwin, including pre-AVX2 x86 with the official baseline runtime
- remove Docker lock generation and install-time Nix side effects, then document the deterministic update workflow
- add CI package smoke coverage and a discriminating Ivy Bridge compatibility gate

## Validation

- bunx tsc --noEmit
- bun run check .
- bun test --isolate --timeout=10000 --max-concurrency=4: 1,729 passed, 4 skipped, 0 failed
- normal compiled release smoke
- deterministic bun run update-nix
- nix flake show --all-systems
- nix flake check
- nix build .#backlog-md, including installed CLI, MCP, browser asset, and older-x86 checks
- nix run .# -- --version

## Known limitation

QEMU user mode crashes Bun JSC when JIT is enabled, so only the emulated Ivy Bridge validation disables JIT. Native installed-package smoke keeps JIT enabled. The gate first proves that the same CPU model rejects normal AVX2 Bun, then verifies the official baseline runtime.

Closes #801


MrLesk:
@codex review

MrLesk:
@codex review

MrLesk:
@codex review

PR #808, opened 2026-07-23
BACK-548 - Add quick toggles for hiding empty board columns (TUI hotkey + browser board button)

janosmiko:
## Summary

Adds quick toggles for the existing `hideEmptyColumns` setting so users can flip it from the surfaces where they actually look at the board, instead of only from the browser Settings page (shipped in BACK-522).

- **TUI**: `H` (Shift-h) hotkey on the board view filters empty status columns at the render boundary via a pure `filterVisibleColumns(data, hideEmptyColumns, isMoving)` helper. Move mode always sees all columns (mirrors the web `isDragging` bypass). The flag persists to `config.yml` with an in-flight guard against key-repeat races and reverts on save failure.
- **Web**: a toggle button in the board toolbar with `aria-pressed` state and optimistic update (reverts on API failure), threaded `App -> BoardPage -> Board`. Empty columns still reappear as drop targets while dragging.
- Both toggles operate on the same shared `hideEmptyColumns` semantics so surfaces stay consistent.

## Drag-and-drop hardening (web)

Re-showing hidden empty columns during a drag needed three fixes to work with native HTML5 drag-and-drop, especially on boards with many statuses:

- **Deferred expansion.** Re-showing columns synchronously inside `dragstart` mutates the board layout, which cancels the native drag in Chromium (cards became undraggable with the toggle on). The expansion is now deferred to the next task with `setTimeout(0)` so the browser commits the drag first; empty columns still reappear as drop targets once the drag is active.
- **Keep the dragged column in view.** Expanding inserts columns to the left of the dragged one, so the board jumped back to the leftmost columns. Its viewport position is captured at `dragstart` and `scrollLeft` restored in a layout effect after expansion, so the source column stays put.
- **Edge auto-scroll.** Native DnD does not auto-scroll, so off-screen columns were unreachable mid-drag. Dragging near the left/right edge now scrolls the board horizontally so any column can be reached.

## Docs

- TUI footer and help popup document the `H` hotkey.
- README Web Interface section mentions both toggles.

## Tests

- `src/test/board-hide-empty-columns.test.ts` - pure-function tests for `filterVisibleColumns` (including move-mode bypass).
- `src/test/web-board-hide-empty-columns.test.tsx` - JSDOM component tests for the browser button (state, label, click handler, absent without handler) plus a regression test that `dragstart` does not re-show hidden columns synchronously (only after the drag commits).

Note: JSDOM has no layout engine, so scroll-preservation and edge auto-scroll were verified manually in a browser rather than in the test suite.

## Verification

- `bunx tsc --noEmit` - clean
- `bun run check .` (Biome) - clean
- `bun test` - green (the hide-empty-columns tests plus full suite)

Note: two pre-existing failures in `tui-task-composer.test.ts` (git-hook cases) reproduce on a clean `main` checkout and are unrelated to this change.


MrLesk:
Alex's agent: The TUI portion of this PR has now shipped in #889, which carries your original commit under your authorship (adapted to the reworked board code) plus Co-authored-by trailers throughout: the existing hideEmptyColumns config key now drives the TUI board, and your Shift+H quick toggle persists it to config. The shortcut lives in the help popup rather than the footer to keep it uncrowded. Your web drag-and-drop findings describe a real defect in already-shipped code and are being triaged separately with your fixes as the starting point. Thank you for the thorough PR. Closing as superseded by #889.

PR #791, opened 2026-07-15
BACK-430 - Create tasks with an intent-first TUI composer

MrLesk:
## Summary

- add a discoverable `N` shortcut and Blessed task composer for title, multiline description, status, type, and priority
- keep the first configured workflow status as the resting default, with Draft available only through explicit status selection
- persist normal tasks and drafts through the canonical creation path with ownership-safe rollback, retry, and Git index preservation
- refresh the board once after success, focus visible tasks, and explain when drafts or filtered tasks cannot be focused

## User experience

The composer supports explicit Create and Cancel actions, preserves entered values after validation or persistence failures, reflows between full and compact layouts, and allows Escape cancellation from every focusable control. An unfiltered empty board can now open normally so the first task can be created from the TUI.

## Validation

- 64 focused tests passed across the composer, help, empty-board loading, auto-commit, CLI commit behavior, atomic creation, and Git operations
- full suite passed with 1,734 tests, 4 intentional interactive skips, and 0 failures
- `bunx tsc --noEmit`
- `bun run check .`
- `bun run build`
- `git diff --check`
- rendered PTY QA at 100x30, 80x24, and 50x18 covered discovery, entry, selectors, creation, cancellation, retry, focus, scrolling, and live resize

Closes #579

PR #809, opened 2026-07-23
BACK-555 - Fix TUI task composer editing bugs and add Ctrl+W word delete

janosmiko:
## Summary

Fixes editing bugs in the TUI task composer (shipped in #791) and brings it to BACK-430 first-slice parity. All four bugs trace to `neo-neo-bblessed` defects the composer did not work around, verified against the installed library.

## Bugs fixed

| Bug | Root cause | Fix |
|-----|-----------|-----|
| Backspace was a no-op in **Description** | `textarea.ts:449` backspace branch is empty under `fullUnicode`, which `createScreen` always sets | Composer owns deletion + render |
| Backspace looked **frozen** in Title | `textbox.ts:117-122` deletes then returns before the trailing `screen.render()` | `ignoreKeys` on the textbox + composer owns deletion + render |
| **Status picker preselected "Draft"** | `list.ts:47` hardcodes `selected = 0` and ignores the `selected` option, so the picker always opened on index 0 (Draft) | `picker.select(selectedIndex)` in shared `filter-popup.ts` |
| **j/k didn't navigate pickers** | `list.ts:119/124` bind j/k only under `vi`, which `filter-popup` never set | `vi: true` in shared `filter-popup.ts` |

The `filter-popup.ts` fixes also correct the board's own filter popups, which had the same preselection and j/k gaps.

## Added

- **Ctrl+W** deletes the previous word in Title and Description.

## Parity

- Resting Status now resolves via `getDefaultCreateStatus` (exported from `task-wizard.ts`), matching the CLI wizard's canonical "To Do" instead of a literal `statuses[0]`.

## Tests

- Pure helpers `deleteLastChar` / `deleteLastWord`.
- Resting-status resolves to canonical "To Do" when present but not first.
- Rendered-harness behavioral tests: Backspace and Ctrl+W edit both Title and Description; the status picker preselects the current value so Enter keeps it (returns "To Do", not "Draft").

## Verification

- `bunx tsc --noEmit` clean
- `bun run check .` clean
- `bun test`: 1781 pass / 4 skip / 2 fail — the two failures (`freezes selected task bytes`, `runs post-commit hooks against the real index`) are pre-existing git-hook tests that also fail on a clean `main` checkout, unrelated to this change.

Note: j/k navigation is a TUI-only interaction the test harness can't drive (list nav binds on the raw `keypress` event); verified manually.

### DIFF
--- a/backlog/tasks/back-555 - Fix-TUI-task-composer-editing-bugs-and-add-CtrlW-word-delete.md
+++ b/backlog/tasks/back-555 - Fix-TUI-task-composer-editing-bugs-and-add-CtrlW-word-delete.md
@@ -1,0 +1,51 @@
+---
+id: BACK-555
+title: Fix TUI task composer editing bugs and add Ctrl+W word delete
+status: In Progress
+assignee:
+  - '@claude'
+created_date: '2026-07-23 05:08'
+updated_date: '2026-07-23 05:20'
+labels:
+  - tui
+  - bug
+dependencies: []
+priority: medium
+type: bug
+ordinal: 200000
+---
+
+## Description
+
+<!-- SECTION:DESCRIPTION:BEGIN -->
+The shipped TUI task composer (PR #791) has editing bugs from unhandled neo-neo-bblessed defects and diverges from the BACK-430 first-slice spec. Fix backspace in Title and Description, stop the Status picker preselecting Draft, enable j/k in pickers, align resting status with the CLI wizard, and add Ctrl+W word deletion in text inputs.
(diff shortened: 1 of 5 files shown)


MrLesk:
Alex's agent: Both fixes from this PR have now shipped. The status picker preselection fix landed in #833, and the vi-style j/k navigation for the filter popups landed in #888, which carries your original commit under your authorship (adapted to the reworked popup code, with the blessed vi flag replaced by explicit j/k bindings so it does not also bind l/q/g/G). Thank you for the report and the fixes. Closing as superseded by #888.

PR #814, opened 2026-07-24
BACK-546 - Add dependency readiness guidance to TUI and browser

cottrell:
## What changed

- Adds canonical dependency-readiness semantics and guidance across the CLI, browser, and interactive TUI.
- Keeps MCP changes as adapter parity for the CLI-defined behavior.
- Resolves readiness against the full task graph without changing task ordering or state.

## Why

Users can identify work that is ready to start and understand what blocks a task without restoring the abandoned derived-sequence model.

## Validation

- `bun test`
- `bunx tsc --noEmit`
- `bun run check .`
- `bun run build`

Recommended merge strategy: squash merge.

### DIFF
--- a/backlog/tasks/back-546 - Add-dependency-readiness-guidance-to-TUI-and-browser.md
+++ b/backlog/tasks/back-546 - Add-dependency-readiness-guidance-to-TUI-and-browser.md
@@ -1,10 +1,11 @@
 ---
 id: BACK-546
 title: Add dependency readiness guidance to TUI and browser
-status: To Do
+status: Done
 assignee:
-  - '@alex-agent'
+  - '@codex'
 created_date: '2026-07-13 16:06'
+updated_date: '2026-08-02 20:31'
 labels:
   - tui
   - web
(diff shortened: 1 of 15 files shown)


MrLesk:
Alex's Agent: Both changes in this draft look useful, but they should be reviewed and merged as two focused PRs.

Please split them as follows:

1. `BACK-546 - Add dependency readiness guidance to TUI and browser`: keep the dependency-readiness semantics and the shared/CLI/TUI/browser guidance that belongs to this existing task. Any MCP changes should remain adapter parity, not define the feature.
2. A new task, `Add ready filter toggle to Kanban board in Web UI and interactive TUI`: move the Web “Ready only” toggle and URL state, the TUI `R` shortcut, and their related tests there. Create this task with `backlog task create` so it receives a fresh ID. Do not reuse `BACK-556`, which conflicts with another task.

Please link each PR to exactly one task and update the titles and bodies accordingly. Both features look worthwhile; splitting them will make review and verification clearer. Please leave this PR as draft until the split is ready.

cottrell:
Alex's Agent: The first split is complete in this draft: it now contains only `BACK-546`, rebuilt from current `main` and validated.

The ready-only Kanban filter is tracked separately as `BACK-565`, but it depends on the shared readiness behavior introduced here. Because this PR's head is in a fork, GitHub cannot use that fork branch as the base of a second upstream stacked PR without an upstream branch. To keep the requested reviews independent, the `BACK-565` PR will be rebased onto `main` and opened after `BACK-546` merges.

MrLesk:
Alex's agent: This feature is now on main via #873, built on your branch — your core commit landed verbatim with authorship preserved, and the takeover adapted it across the intervening changes (completed-corpus loading now costs ~112ms instead of seconds, canonical dependency resolution, fail-closed unknown dependencies, plus the review hardening rounds). The Ready-only toggle and TUI shortcut half stays split out as discussed, tracked in BACK-601's family. Thanks for the contribution and the clean split.

PR #817, opened 2026-07-25
BACK-562 - Honor BROWSER for devcontainer browser launch

### LINKED ISSUE #815: [Bug]: `backlog browser` fails to open from VS Code devcontainer
## Description

Running `backlog browser` inside a VS Code devcontainer starts the web server, but automatic browser opening fails because Backlog.md invokes `xdg-open` directly on Linux.

VS Code devcontainers provide the `BROWSER` environment variable as a helper that forwards URLs to the host machine's browser. Other development tools that honor `BROWSER` therefore open URLs correctly from inside the container. Backlog.md does not use this mechanism.

## Environment

- Backlog.md: 1.48.0
- Environment: VS Code devcontainer on Linux
- `BROWSER` is set to VS Code's host-browser forwarding helper, for example:

  ```text
  /vscode/vscode-server/bin/linux-x64/<commit>/bin/helpers/browser.sh
  ```

- `xdg-open` is unavailable or cannot open a browser in the container

## Steps to reproduce

1. Open a project in a VS Code devcontainer.
2. Confirm that the container's browser helper works:

   ```sh
   "$BROWSER" https://example.com
   ```

3. Run:

   ```sh
   backlog browser
   ```

4. Observe that Backlog.md attempts to execute `xdg-open` and reports that it could not open the browser automatically.

## Expected behavior

`backlog browser` should open the Web UI in the host browser using the environment-provided browser command when `BROWSER` is set. On Linux, it should fall back to `xdg-open` only when no browser override is configured.

## Actual behavior

Backlog.md unconditionally invokes `xdg-open` on Linux. This bypasses VS Code's devcontainer URL-forwarding helper and fails in containers without a desktop environment or `xdg-open` installation.

## Suggested fix

Honor the conventional `BROWSER` environment variable before using the platform default. For example:

1. If `BROWSER` is non-empty, invoke it with the URL.
2. Otherwise use the existing platform behavior:
   - macOS: `open`
   - Windows: `cmd /c start`
   - Linux: `xdg-open`

The implementation should safely handle a browser command containing arguments, if that is intended to be supported.

An explicit CLI or configuration option for the browser command would also be useful, but honoring `BROWSER` would solve the standard devcontainer case.

## Workarounds

The port-forwarded URL can be opened manually - and the error message prevented via the `--no-open` option.

Alternatively, install an `xdg-open` shim in the container that delegates to `$BROWSER`.

shixi-li:
## Summary

Honor a non-empty BROWSER executable when Backlog.md opens the web UI from either the CLI or server. The browser executable and URL remain separate argv values; quoted executable paths are accepted; macOS, Windows, and Linux retain their existing fallbacks when no usable override is configured. Failure output continues to give manual-open guidance.

## Related Issue or Task

- Closes #815
- Task: BACK-562 — https://github.com/MrLesk/Backlog.md/blob/main/backlog/tasks/back-562%20-%20Honor-BROWSER-for-devcontainer-browser-launch.md

## Validation

- Focused browser-launch and CLI browser tests
- TypeScript typecheck, Biome, and build
- CI-equivalent isolated full suite on current origin/main and this PR head
- GitHub CI includes the required Windows test and compiled-binary paths.


shixi-li:
Quick follow-up on this PR: the current head is mergeable and the focused browser-opening tests, type checks, formatting checks, and build are documented in the PR. The fork CI run is still `action_required` and has not executed. Could a maintainer approve the workflow and take a look when convenient? Happy to address any feedback.

shixi-li:
The change landed on `main` via commit `83fca840bd` ("BACK-562 - Honor BROWSER for devcontainer browser launch (#817)"). Thanks for taking it in!

PR #811, opened 2026-07-23
BACK-555 - Default web server hostname to 127.0.0.1

cbeaulieu-gt:
## Summary
Fixes a security issue: `backlog browser`'s embedded web server bound to `0.0.0.0` (all network interfaces) with no authentication, exposing the full task API to anyone on the same LAN/VPN/Wi-Fi segment. The startup log printed `http://localhost:${port}`, which was misleading since the server wasn't actually localhost-restricted.

- `BacklogServer.start()` now defaults `hostname` to `127.0.0.1` in `serveOptions` passed to `Bun.serve()`.
- `start()` accepts a third, optional `hostname` argument so callers can opt into a different bind address.
- The `backlog browser` CLI command gains an opt-in `--host <address>` flag, threaded through to `start()`, for anyone who intentionally wants LAN access.

## Related Issue or Task
closes #810
Task: `backlog/tasks/back-555 - Default-web-server-hostname-to-127.0.0.1.md` (BACK-555)

## Task Checklist
- [x] I have created a corresponding task in `backlog/tasks/`
- [x] The task has clear acceptance criteria
- [x] I have added an implementation plan to the task
- [x] All acceptance criteria in the task are marked as completed

## Testing
Two new frozen-first tests were written before the fix (test-first split) and now pass:
- `src/test/server-hostname.test.ts` — `BacklogServer.start()` binds `127.0.0.1` by default; honors an explicit `hostname` override.
- `src/test/cli-browser-host.test.ts` — `browser --host <address>` is accepted by the CLI and the server boots successfully.

Full verification (mirrors CI):
- `bunx tsc --noEmit` — clean.
- `bunx biome lint .` — clean (341 files checked, no fixes needed).
- `bun test --isolate --timeout=10000 --max-concurrency=4` (full suite, CI flags) — **1770 pass / 15 skip / 1 fail**. The 1 failure (`installClaudeAgent > writes the project-manager-backlog.md file with correct content`) is a pre-existing baseline failure, confirmed identical on unmodified `main` via stash-and-rerun — unrelated to this change.

Diff is minimal: 2 implementation files (4 production lines changed), 2 new test files, 1 task record.

---
🤖 _Generated by Claude Code on behalf of @cbeaulieu-gt_

PR #826, opened 2026-07-30
BACK-560 - Bind the browser server to loopback only

### LINKED ISSUE #810: Web UI (`backlog browser`) binds to 0.0.0.0 with no authentication — full task data exposed on LAN
## Summary

`backlog browser` (the local web GUI/server, `src/server/index.ts`) binds to `0.0.0.0` (all network interfaces) with no authentication, and there is currently no `--host`/`--hostname` flag or config option to restrict it to localhost.

## Impact

Anyone on the same network segment as the machine running `backlog browser` — office Wi-Fi, shared VPN subnet, coworking space, etc. — can reach the full task board and its API with zero credentials:

```
curl http://<host-ip>:6420/api/tasks
```

This includes read access to all tasks/documents/decisions/config, and write access via the same unauthenticated endpoints (create/edit/delete/complete tasks). The startup log prints `http://localhost:${port}`, which is misleading since the server is not actually localhost-restricted — this could give a false sense of safety.

## Where this lives

`serveOptions` passed to `Bun.serve()` (around `src/server/index.ts:296`) has no `hostname` key. Per Bun's docs, `Bun.serve()` defaults to `0.0.0.0` when `hostname` is omitted.

## Proposed fix

- Default `hostname` to `127.0.0.1` in `serveOptions`.
- Add an opt-in `--host <address>` (or `--hostname`) CLI flag / config field for anyone who intentionally wants LAN access, mirroring the pattern MCP's HTTP transport already uses (per #355 — MCP HTTP/SSE is restricted to `127.0.0.1`/`::1` by default).

## Related prior art

#701 ("authenticated multi-user hosting") proposed a broader `--host`-plus-OIDC-auth feature and was closed without any comments/review — no stated reason, so it's unclear whether the maintainers are against a `--host` flag in principle or the PR was just too large in scope. This issue is intentionally scoped much smaller: just default-safe binding + an opt-in flag, no auth system.

Happy to open a PR for the scoped version (default `127.0.0.1` + `--host` flag) if that's welcome.

---
🤖 _Generated by Claude Code on behalf of @cbeaulieu-gt_

MrLesk:
## Summary

- Bind the local browser server explicitly to `127.0.0.1`.
- Use the same loopback interface for port probing, startup output, and automatic browser opening.
- Document `backlog browser` as local-machine only and keep `--no-open` behavior unchanged.
- Do not add a public host override or unauthenticated LAN mode.

## Related issue and contribution

Fixes #810.

This takes over the safe loopback-binding portion of PR #811 after the maintainer decision to keep the browser server loopback-only. PR #811 could not be merged directly because its BACK-555 task ID collides with current main and its port probe still checked a different interface from the server binding.

## Verification

- Focused browser binding and port tests: 12 passed
- Broader server/browser tests: 74 passed
- Full suite: 1,782 passed, 4 skipped, 0 failed
- `bunx tsc --noEmit`
- `bun run check .`
- Live verification: loopback returned HTTP 200 while the tested Wi-Fi and VPN addresses were unreachable

### DIFF
--- a/CLI-INSTRUCTIONS.md
+++ b/CLI-INSTRUCTIONS.md
@@ -253,8 +253,10 @@ Manage task dependencies to express execution order:
 
 | Action      | Example                                              |
 |-------------|------------------------------------------------------|
-| Web interface | `backlog browser` (launches web UI on port 6420) |
+| Web interface | `backlog browser` (launches the local-machine-only web UI on `127.0.0.1:6420`) |
 | Web custom port | `backlog browser --port 8080 --no-open` |
+
+The Web UI listens only on `127.0.0.1`; it is not reachable from other devices on the LAN or VPN.
 
 To keep the Web UI running in the background with auto-start on boot, see [Running Backlog.md as a Service](backlog/docs/doc-003%20-%20Running-Backlog-Browser-as-a-Service.md).
 
--- a/README.md
+++ b/README.md
@@ -184,7 +184,8 @@ Read commands support stable, versioned JSON for scripts and integrations. Use `
 
 ## <img src="./.github/web-interface-256.png" alt="Web Interface" width="28" height="28" align="center"> Web Interface
 
-Launch a local web interface for visual task management:
+Launch a web interface for visual task management on the local machine. The server listens on `127.0.0.1` and is not
+reachable from other devices on the LAN or VPN:
 
 ```bash
 # Start the web server (opens browser automatically)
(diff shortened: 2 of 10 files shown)


longyuxi:
Hi, thanks for the great software. Mind if I ask the reason to not allow specification of the hostname? I think it would be more convenient to give the user a choice to expose to LAN.

MrLesk:
Hey @longyuxi 
The web interface is designed as a GUI only. Imagine it's a desktop app running on your computer. It is not at all hardened against external access and can become a critical vulnerability for regular users.

If someone needs to access Backlog.md from another computer they can use a tunnel `ssh -L 6420:127.0.0.1:6420 alex@192.168.x.x` or via tailscale.

Or maintain their own fork of Backlog.md without security guarantees.

PR #828, opened 2026-07-30
BACK-559 - Eliminate repeated task-corpus scans from browser updates

### LINKED ISSUE #807: [Bug]: Web UI seems significantly slower than previous version & seems to tax my machine.
**Describe the bug**
When I run `backlog browser` and open the window, it loads fine, but whenever I make a change (i.e. dragging a task from one column to the other), it takes a ~4-5 seconds for it to show up, and I can hear my computer's fan spinning up. This didn't happen on the previous version.

**To Reproduce**
1. Upgrade to version 1.48.0
2. Open a project with ~450 tasks (most of which have been moved to the "completed" folder so no longer show up on the board)
3. Move a task from one column to the next

**Expected behavior**
I expect the UI to update right away and for it not to cause my computer a lot of stress.

**Environment**
- OS: Windows 11, Powershell 7, with an i7-14700KF Processor
- Node version: 26.2.0

MrLesk:
## Summary

- resolve browser task mutations once and persist the resolved task without reloading the full active and completed corpus
- reuse one active and completed snapshot for duplicate repair and reorder work while preserving fail-closed identity checks
- return every persisted reorder change and apply it through the existing stale-response guard, keeping a forced two-task rebalance correct when WebSocket delivery is unavailable
- coalesce multi-task reorder publication to one WebSocket reconciliation and ignore mutation responses superseded by newer reconciled state

## Verification

- `bun test`: 1,789 passed, 4 interactive tests skipped, 0 failed, with 7,590 expectation calls across 200 files
- focused server and rendered App correction suite: 37 passed, 0 failed
- `bunx tsc --noEmit`
- `bun run check .`: 340 files checked with no errors
- `git diff --check`

## Performance

On alternating fresh same-machine macOS arm64 Git fixture copies with 20 active and 430 completed tasks, each variant ran 12 measured samples. Every sample forced the same two-task ordinal rebalance. End-to-end completion included the mutation response and every surviving App refresh request: statuses, config, search, milestones, archived milestones, and duplicate preview. Current `origin/main` included its foreground refresh plus both WebSocket reconciliations. This change returned both changed tasks immediately and completed one WebSocket reconciliation.

- current `origin/main`: 1,285.543 ms median, 1,366.933 ms p95
- this change: 108.588 ms median, 241.318 ms p95
- median reduction: 91.6%

No durable benchmark framework or fixture is added.

Fixes #807

### DIFF
--- a/backlog/tasks/back-559 - Eliminate-repeated-task-corpus-scans-from-browser-updates.md
+++ b/backlog/tasks/back-559 - Eliminate-repeated-task-corpus-scans-from-browser-updates.md
@@ -5,7 +5,7 @@ status: Done
 assignee:
   - '@codex'
 created_date: '2026-07-30 17:12'
-updated_date: '2026-07-30 18:07'
+updated_date: '2026-07-30 18:33'
 labels:
   - web-ui
   - performance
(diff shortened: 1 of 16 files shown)


MrLesk:
Alex's Agent: Superseded by PR #830, “BACK-559 - Make Core the sole browser task boundary,” which is now merged. PR #830 reimplemented BACK-559 from current main, makes Core the sole browser task boundary, preserves fail-closed identity and exact-path lifecycle behavior, and includes the performance improvements. Closing this earlier implementation so only the completed fix remains active.

PR #850, opened 2026-08-06
Allow clearing defaultEditor via empty value in config set and init

### LINKED ISSUE #844: [Bug]: `defaultEditor` cannot be cleared — `config set ""` is rejected and `init --default-editor ""` is silently ignored
## Summary

There is no way, through the CLI, to configure a project to have **no** editor. Both routes fail,
one of them silently. The only working fix is hand-editing `backlog/config.yml`.

This matters for unattended/agent use: the shipped default is `code --wait`, and an editor that
waits blocks a non-interactive process indefinitely.

## Reproduction

**Route 1 — `config set` rejects the empty value:**

```sh
$ backlog config set defaultEditor ""
Editor command not found:
Please ensure the editor is installed and available in your PATH
```

The value is validated as an executable before it is stored, so the one value that means *"do not
open an editor"* is the one value that cannot be stored.

**Route 2 — `init` accepts the flag and ignores it:**

```sh
$ backlog init "P" --defaults --default-editor "" ...
...
│    Default editor: code --wait   │      <- the flag had no effect
$ grep default_editor backlog/config.yml
default_editor: "code --wait"
```

The empty string is treated as "not supplied" and the default is applied. The summary does print
the real value, so it is not *entirely* silent — but nothing indicates the flag was discarded.

## Workarounds, for anyone who lands here

- Hand-edit `backlog/config.yml` to `default_editor: ""` — the CLI reads this fine for every
  non-editor command.
- Or `backlog config set defaultEditor true` — `true(1)` is a real binary that exits 0 instantly,
  so an accidental editor launch returns immediately instead of hanging.

## Suggested fix

Treat an explicitly-empty value as a valid "no editor" setting in both paths, rather than as a
missing one — and skip the executable-exists validation when the value is empty.

## Environment

backlog.md 1.49.3 (homebrew) · macOS arm64.

iRonin:
## Summary

There is no way through the CLI to configure a project to have **no** editor:

- `backlog config set defaultEditor ""` is rejected with `Editor command not found` — the value is validated as an executable before storing, so the one value meaning "do not open an editor" cannot be stored.
- `backlog init --default-editor "" ...` silently discards the flag because the empty string is falsy and falls through to the existing-config/`EDITOR`/`VISUAL` fallback chain.

This matters for unattended/agent use, where a blocking editor default such as `code --wait` hangs a non-interactive process.

This PR treats an explicitly empty value as a valid "no editor" setting in both paths:

- `config set defaultEditor` skips the executable-exists validation when the value is empty and stores it (serialization omits the empty value, so `default_editor` is removed from `config.yml`).
- `init --default-editor ""` now counts as "flag provided" (using `??` instead of `||` so the empty string no longer falls through to the env/existing-config fallback). The existing clearing logic in `initializeProject` (`delete config.defaultEditor` when an override is present but empty) then applies, which also makes re-init with `--default-editor ""` clear a previously configured editor.

## Related Issue or Task

Fixes #844

> **📋 Important:** Please discuss the change in an issue before opening a PR.
> - If no issue exists, open one first so maintainers and contributors can agree that the proposal fits the current project scope before implementation.
> - All PRs must have an associated task in the backlog.
> - If no task exists, create one first using: `backlog task create "Your task title"`
> - Follow the [task guidelines](../src/guidelines/agent-guidelines.md) when creating tasks
> - Tasks should be atomic, testable, and well-defined with clear acceptance criteria

## Task Checklist

- [ ] I have created a corresponding task in `backlog/tasks/` — _no task file added; happy for the maintainer to allocate a task ID at review_
- [ ] The task has clear acceptance criteria
- [ ] I have added an implementation plan to the task
- [ ] All acceptance criteria in the task are marked as completed

## Testing

Added three regression tests in `src/test/config-commands.test.ts` (all fail on `main`, pass with this change):

- `clears defaultEditor via config set with an explicitly empty value` — previously exited 1 with `Editor command not found`; now exits 0 and the reloaded config has no editor.
- `honors an explicitly empty --default-editor flag during init instead of discarding it` — with `EDITOR` set to a sentinel value, asserts the initialized config has no editor (a discarded empty flag would have leaked the sentinel into the config).
- `clears a configured editor when re-initializing with an explicitly empty --default-editor` — verifies the `EDITOR` env fallback still applies when the flag is absent, and that re-init with `--default-editor ""` clears it.

- `bun test src/test/config-commands.test.ts`: 16 pass, 0 fail.
- `bunx tsc --noEmit`: clean.
- `bunx biome check .`: clean.
- Full `bun test` (1897 tests, 213 files, macOS arm64): all remaining failures are pre-existing flaky ~5s timeouts in unrelated git/remote/config-watcher tests — reproduced identically on a clean `main` worktree, with run-to-run variance on the same tree.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -824,7 +824,7 @@ addHelpSchema(program.command("init [projectName]"), {
 					options.branchDays ||
 					options.bypassGitHooks ||
 					options.zeroPaddedIds ||
-					options.defaultEditor ||
+					options.defaultEditor !== undefined ||
 					options.webPort ||
 					options.autoOpenBrowser ||
 					options.installClaudeAgent ||
@@ -1029,11 +1029,8 @@ addHelpSchema(program.command("init [projectName]"), {
 					const paddingValue = parseNumber(options.zeroPaddedIds, result.zeroPaddedIds ?? 0);
 					result.zeroPaddedIds = paddingValue > 0 ? paddingValue : undefined;
 					result.defaultEditor =
-						options.defaultEditor ||
-						existingConfig?.defaultEditor ||
-						process.env.EDITOR ||
-						process.env.VISUAL ||
-						undefined;
+						options.defaultEditor ??
+						(existingConfig?.defaultEditor || process.env.EDITOR || process.env.VISUAL || undefined);
 					result.defaultPort = parseNumber(options.webPort, result.defaultPort ?? 6420);
 					result.autoOpenBrowser = parseBoolean(options.autoOpenBrowser, result.autoOpenBrowser ?? true);
 					return result;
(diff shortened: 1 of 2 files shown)


iRonin:
Closing for now per our contribution workflow — we have implemented this fix in our fork (iRonin/Backlog.md, branch fix/allow-empty-default-editor) and are happy to open a PR whenever you'd like one. Details on #844.

MrLesk:
Alex's agent: The fix from this branch was merged into main via #855 with your commit authorship preserved, and #844 is closed. Thanks for the contribution and for the quality of the whole report batch.

PR #852, opened 2026-08-06
Serialize the task edit read-modify-write with a per-task lock

### LINKED ISSUE #843: [Bug]: `task edit` loses concurrent writes silently — #565's lock was never applied to the edit path
## Summary

Two `backlog task edit` calls on the same task, running concurrently, lose one of the two writes.
**Both processes exit 0 and both print `Updated task <id>`.** There is no error at either call site.

#565 ("Harden task ID locking for concurrent create/promote/demote operations") added a
library-backed lock, but it serialises **ID allocation at create time**. The read-modify-write in
`task edit` is not covered by it. So this is a missed call site rather than a request to add
locking — the lock, the `USE_GLOBAL_TASK_ID_LOCK` escape hatch, and
`scripts/smoke-parallel-task-locking.sh` are all already in the tree.

## Reproduction

```sh
backlog task create "Probe"                 # -> task-1
backlog task edit task-1 -s "In Progress"  &
backlog task edit task-1 --notes "writer B" &
wait
grep -c "^status: In Progress" backlog/tasks/task-1*.md   # often 0
grep -c "writer B"             backlog/tasks/task-1*.md   # often 0
```

One of the two is missing. Which one varies.

## Measurements

**Loss rate against the stagger between the two writes** (12 trials per row):

| stagger | trials losing a write |
|---|---|
| simultaneous | 12 / 12 |
| 5 ms | 9 / 12 |
| 10 ms | 11 / 12 |
| 20 ms | 3 / 12 |
| 30 ms | 1 / 12 |
| 50 ms | 1 / 12 |

So the vulnerable window is roughly **20–30 ms** — the read→write gap, not the ~200 ms wall time
of the whole command.

**Concurrent claims** — six processes assigning themselves to one task at once:

```
all 6 exit 0
all 6 print "Updated task PA-3"
1 assignee survives
```

## Controls

Both included because a race report without them is just a flaky test:

- **Negative control — different tasks are safe.** Six concurrent `task edit --notes` calls on six
  *different* tasks: 6/6 survived. So the collision is per-file, not ambient flakiness.
- **Localising control — a mutex fixes it.** Wrapping the identical calls in a `mkdir`-based mutex
  (nothing else changed) takes the same 10-trial test from **10/10 losing a write to 0/10**. That
  places the defect in the unserialised read-modify-write and nowhere else.

## Why it matters here

Running an agent fleet, several sessions write to one board. The common shape is not two agents
fighting — it is a lead setting `status` at the moment the worker writes its final `--notes`. Both
are legitimate, and one silently loses. Losing implementation notes loses actual work, and nothing
surfaces it.

With `autoCommit: true` the loss still happens **and is usually still silent.** Re-measured on a
git-backed board, 10 trials of the same two-writer race: **9 silent, 1** surfaced as
`fatal: Unable to create '.git/index.lock'`. Different-task collisions: 0/6 surfaced. The losing
writer also leaves no commit, so git history does not reveal the loss either.

What `autoCommit` *does* guarantee, since #796, is that a commit contains only its own task's
file — the safety property holds; the visibility one does not.

## This is a divergence from the manifesto, not a feature request

The manifesto already commits to the mitigation that is missing here:

- Design principle 6 — *"Deterministic and reversible where possible. Prefer preview, explicit
  confirmation, **atomic writes**, and useful recovery evidence."*
- Named risk 3, "Plain-text corruption" — *"Friendly files are still structured state.
  Mitigation: semantic commands, validation, **atomic writes**, ambiguity checks, and
  deterministic serialization."*
- Principle 4 — *"Review before consequence"* — and principle 5, *"Fail closed when identity is
  ambiguous. Never guess which task a read or mutation should target."* Here the write does not
  fail closed; it succeeds twice and keeps one result.

And `AGENTS.md`, under simplicity-first rules: *"Keep behavior consistent across similar stores
(defaults, parse errors, **locking**). Divergence requires a clear reason."* Allocation is locked
and edit is not, which is that divergence.

The user-visible cost is the one the manifesto cares about most — *"preserve the record"* is the
last step of the core loop, and this is the step that loses it. Implementation notes are written
once, at the end of a task, by whoever did the work.

## Suggested fix

Extend the existing lock to cover the edit read-modify-write, keeping the critical section to the
read + mutate + write as #565 did for allocation.

There is a worked precedent in a sibling project — `claude-task-master`
([issue #1567](https://github.com/eyaltoledano/claude-task-master/issues/1567)) hit the identical
bug and fixed it with `proper-lockfile` plus atomic temp-write-and-rename, and crucially
**re-reading the file inside the lock** so the mutation is applied to current state rather than to
a snapshot taken before the wait.

`scripts/smoke-parallel-task-locking.sh` looks like it would extend to cover `edit` directly.

## Environment

backlog.md 1.49.3 (homebrew) · macOS arm64 · `filesystem_only` both true and false, same result.

iRonin:
## Summary

`backlog task edit` performs an unlocked read-modify-write: the task file is parsed into a snapshot, mutated in memory, and the whole file is serialised back. Two concurrent edits of the same task lose one write — **both processes exit 0 and both print `Updated task X`** (measured in #843: 12/12 writes lost when simultaneous; a ~25 ms vulnerable window).

#565 added the `proper-lockfile`-backed lock, scoped to ID allocation at create/promote/demote time. This PR finishes that work for the edit path: a **per-task lock around the whole read → mutate → write** in `updateTaskFromInput`, which the CLI (`task edit`), the MCP server (`update_task` via `editTaskOrDraft`), and the web server (PUT task) all funnel through — one fix point covers every front door.

Design notes:

- New `FileSystem.withTaskLock(task, fn)` beside `withCreateLock`, sharing its mechanics (extracted into a common `withLockTarget`), its lock location (git common dir when present, `backlog/.locks` otherwise — worktree-safe), its timeout/stale defaults, and the `USE_GLOBAL_TASK_ID_LOCK=false` escape hatch. Lock contention surfaces as a user-facing `ETASKLOCK` error after the same 30 s default, matching `ECREATELOCK`.
- **The lock target is the task file itself**, not the shared backlog dir: proper-lockfile tracks one lock per target file in-process, so a shared target with per-task `lockfilePath`s corrupts concurrent locks held by one process (observed as `ERELEASED` when two edits in a single process race). Per-task targets also mean an edit never contends with the create lock.
- **The snapshot is re-read inside the lock.** A lock around only the write would still apply a waiting writer's mutation to its pre-wait snapshot, so the lost write would survive the lock. One regression test asserts against exactly that implementation.
- The mutation input — not the mutated object — crosses the lock boundary, per the same shape as #565's allocation fix. The write itself is unchanged (still `Bun.write`); temp-write-and-rename would harden against crash mid-write and can be a follow-up if you want it.

Not in this PR — the same defect on other write paths, deliberately kept separate per the small-reviewable-units rule: `reorderTask`/`updateTasksBulk` (board drag; archive link-sanitiser — needs a multi-file story), draft edits (`updateDraftFromInput` → `saveDraft`), and the TUI's raw `$EDITOR` write (a lock would span an editor session). Happy to take any of these as follow-ups.

The manifesto already commits to this mitigation: design principle 6 ("Deterministic and reversible where possible. Prefer preview, explicit confirmation, **atomic writes**, and useful recovery evidence") and named risk 3, "Plain-text corruption" ("Mitigation: semantic commands, validation, **atomic writes**, ambiguity checks, and deterministic serialization"). And AGENTS.md asks to "keep behavior consistent across similar stores (defaults, parse errors, **locking**); divergence requires a clear reason" — allocation is locked and edit was not; this removes that divergence.

## Related Issue or Task

Fixes #843

No `backlog/tasks/` entry is included — I didn't want to claim a BACK- id unilaterally. Happy to add one under whatever id you assign.

## Task Checklist

- [ ] I have created a corresponding task in `backlog/tasks/` — see note above; will add on id assignment
- [ ] The task has clear acceptance criteria — see note above
- [ ] I have added an implementation plan to the task — see note above
- [ ] All acceptance criteria in the task are marked as completed — see note above

## Testing

New `src/test/atomic-task-edit.test.ts` (5 tests, green on this branch):

- **`preserves every change when concurrent edits target the same task`** — 6 writers append distinct labels to one task; asserts all 6 labels are present in the final file. Content assertion, because every racer exits 0 either way. **Red on `main`: 1–2 of 6 labels survive (3/3 runs)**; green with the fix.
- **`applies an edit that waited on the lock on top of the change that held it`** — a writer takes its pre-lock snapshot, blocks while the lock holder writes, then must land on the holder's state. The negative this returns: a lock-without-re-read implementation loses the holder's label and fails.
- Concurrent edits of *different* tasks stay independent (the issue's negative control); lock timeout produces the user-facing `ETASKLOCK` message; the legacy escape hatch bypasses serialisation.

`scripts/smoke-parallel-task-locking.sh` gains **scenario 4: concurrent task editing**, reusing the existing harness and asserting on final file content. One subtlety, documented in the script: the board is pre-filled with 100 tasks, because via `bun src/cli.ts` the ~25 ms race window is otherwise hidden by process-startup variance — without the fillers the scenario passes even on the broken build (verified: 8/8 labels survive pre-fix with 1 task, 1–2/8 with 100). Red on `main` ("Label smoke-1 missing after parallel edits — a concurrent edit was lost"), green with the fix. Scenarios 1–3 unchanged and passing in both states.

Full suite: `bun test` — 1899 tests across 214 files. On macOS the git-branch/fs-watcher CLI-integration tests (e.g. `should merge task status from remote branches`, the ContentStore watcher-publish tests) flake on their 5 s timeouts under load; the failing set varies run to run (2–18 failures), and the same files time out identically on **unmodified** `upstream/main` in isolation, so these are pre-existing (upstream's macOS CI runs only a subset of the suite) and are untouched here. All locking, task-edit, create, MCP, and web tests pass; the new tests pass in 3/3 consecutive full-file runs.

`bunx tsc --noEmit` clean · `bunx biome check .` clean.

### DIFF
--- a/scripts/smoke-parallel-task-locking.sh
+++ b/scripts/smoke-parallel-task-locking.sh
@@ -23,6 +23,7 @@ Runs real parallel smoke tests against the local Backlog.md CLI:
   1. concurrent task creation
   2. concurrent draft promotion
   3. concurrent task demotion
+  4. concurrent task editing (lost-write regression)
 
 Environment overrides:
   BUN_BIN            Bun executable to use (default: bun)
@@ -231,6 +232,42 @@ scenario_parallel_demote() {
 	info "Scenario 3 passed"
 }
 
+scenario_parallel_edit() {
+	local repo_dir="${TMP_ROOT}/edit"
+	local i
+
+	info "Scenario 4: concurrent task editing (${JOB_COUNT} jobs)"
+	init_repo "${repo_dir}"
+
+	pushd "${repo_dir}" >/dev/null
+	run_cli task create "Shared Task" >/dev/null
+	popd >/dev/null
+
+	# All racers exit 0 and print "Updated task" either way — only the final file
+	# content can reveal a lost write, so that is what this scenario asserts on.
+	for i in $(seq 1 "${JOB_COUNT}"); do
+		start_job "${repo_dir}" "edit-${i}" task edit task-1 --add-label "smoke-${i}"
+	done
+	wait_for_jobs
+
+	local task_files=("${repo_dir}"/backlog/tasks/task-1\ -\ *.md)
+	if [[ ${#task_files[@]} -ne 1 ]]; then
+		ls -la "${repo_dir}/backlog/tasks" >&2 || true
+		fail "Expected exactly one task-1 file after parallel edits"
+	fi
+
+	local content
+	content="$(cat "${task_files[0]}")"
+	for i in $(seq 1 "${JOB_COUNT}"); do
+		if ! printf '%s' "${content}" | grep -Eq "smoke-${i}([^0-9]|$)"; then
+			printf '[smoke] Final task file after parallel edits:\n%s\n' "${content}" >&2
+			fail "Label smoke-${i} missing after parallel edits — a concurrent edit was lost"
+		fi
+	done
+
+	info "Scenario 4 passed"
+}
+
 main() {
 	while [[ $# -gt 0 ]]; do
 		case "$1" in
@@ -277,6 +314,7 @@ main() {
 	scenario_parallel_create
 	scenario_parallel_promote
 	scenario_parallel_demote
+	scenario_parallel_edit
 
 	info "All parallel locking smoke tests passed"
 }
--- a/scripts/smoke-parallel-task-locking.sh
+++ b/scripts/smoke-parallel-task-locking.sh
@@ -234,26 +234,36 @@ scenario_parallel_demote() {
 
 scenario_parallel_edit() {
 	local repo_dir="${TMP_ROOT}/edit"
+	local filler_count=100
+	local shared_id="task-$((filler_count + 1))"
 	local i
 
 	info "Scenario 4: concurrent task editing (${JOB_COUNT} jobs)"
 	init_repo "${repo_dir}"
 
+	# The race window of one edit is its ~25ms read-modify-write, which bun startup
+	# variance otherwise hides: one racer has usually finished before the next starts
+	# parsing. A board full of filler tasks stretches every edit's read phase so the
+	# windows genuinely overlap. Without the fillers this scenario passes even on the
+	# broken build (verified: 8/8 labels survive pre-fix with 1 task, 1-2/8 with 100).
 	pushd "${repo_dir}" >/dev/null
+	for i in $(seq 1 "${filler_count}"); do
+		run_cli task create "Filler ${i}" >/dev/null
+	done
 	run_cli task create "Shared Task" >/dev/null
 	popd >/dev/null
 
 	# All racers exit 0 and print "Updated task" either way — only the final file
 	# content can reveal a lost write, so that is what this scenario asserts on.
 	for i in $(seq 1 "${JOB_COUNT}"); do
-		start_job "${repo_dir}" "edit-${i}" task edit task-1 --add-label "smoke-${i}"
+		start_job "${repo_dir}" "edit-${i}" task edit "${shared_id}" --add-label "smoke-${i}"
 	done
 	wait_for_jobs
 
-	local task_files=("${repo_dir}"/backlog/tasks/task-1\ -\ *.md)
+	local task_files=("${repo_dir}"/backlog/tasks/${shared_id}\ -\ *.md)
 	if [[ ${#task_files[@]} -ne 1 ]]; then
 		ls -la "${repo_dir}/backlog/tasks" >&2 || true
-		fail "Expected exactly one task-1 file after parallel edits"
+		fail "Expected exactly one ${shared_id} file after parallel edits"
 	fi
 
 	local content
--- a/src/core/backlog.ts
+++ b/src/core/backlog.ts
@@ -2033,16 +2033,26 @@ export class Core {
 			return await this.demoteTaskWithUpdates(task, input, autoCommit);
 		}
 
-		const { mutated } = await this.applyTaskUpdateInput(task, input, async (status) =>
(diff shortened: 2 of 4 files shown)


iRonin:
Closing this for now — it was opened ahead of our own verification, and I would rather withdraw it than leave a concurrency patch in front of you that we have not finished testing.

Specifically: the acceptance test for this change is a **concurrency race harness**, not the unit suite. The suite passes with the defect present — 1,891 tests, none of which covers a lost update — so a green run proves nothing here. Our measured baseline is 49 of 60 writes lost across six concurrent `task edit` calls; the patch has not yet been run against that harness.

The commit history also contains internal recovery commits from our side that shouldn't be in front of you.

The analysis in the description stands and I'll keep it: four write paths share the `updateTaskFromInput` funnel, and a lock alone is insufficient because the mutation is applied to a snapshot taken before the wait — the re-read has to happen inside the critical section.

We're running the fix in our own fork meanwhile. Happy to open a clean PR whenever you'd like one — details on #843.

MrLesk:
Alex's agent: Concurrent task edits are now protected on main via #860. The maintainer opted for fail-fast semantics rather than wait-and-re-read: on contention the second writer exits non-zero with a clear error (web returns 409, MCP an operation error) and the caller decides whether to retry. Your lock plumbing shape (withTaskLock beside withCreateLock, task-file lock target) and your test harness design — including the filler-task trick to widen the race window — carried over with credit in the task record; the semantics deliberately differ. Thanks again for the report quality: the measured loss rates and the withdrawn-as-unverified honesty made this a straightforward take-up.

PR #857, opened 2026-08-07
BACK-578 - Stop the Rosetta sysctl probe from leaking stderr

### LINKED ISSUE #803: [Bug]: Caught Rosetta sysctl probe leaks stderr in restricted macOS shells
**Describe the bug**

Backlog.md's macOS launcher performs a best-effort Rosetta probe with `execFileSync("/usr/sbin/sysctl", ["-in", "sysctl.proc_translated"])`. When a restricted shell denies that probe, `isRosettaTranslated()` catches the exception and Backlog continues successfully, but the child process's stderr has already leaked into every command's output.

Observed prefix before otherwise successful commands:

```text
sysctl: sysctl fmt -1 1024 1: Operation not permitted
```

This makes successful CLI output look like a failure and complicates agent/CI parsing.

**To Reproduce**

1. Install Backlog.md 1.48.0 through Homebrew on macOS.
2. Run it in a sandbox that permits the CLI but denies `sysctl.proc_translated`.
3. Run `backlog --version` or `backlog instructions overview`.
4. Observe the `sysctl` error on stderr followed by successful Backlog output and exit status 0.

The installed `resolveBinary.cjs` calls `execFileSync` without a `stdio` option, so the probe inherits stderr even though its exception is caught.

**Expected behavior**

A failed optional Rosetta probe should be silent. For example, capture or ignore the probe's stderr while retaining the existing `catch { return false; }` fallback.

**Environment**

- OS: macOS 26.5.2, arm64
- Node version: 26.5.0
- Backlog.md: 1.48.0, Homebrew `backlog-md`

**Additional context**

This is distinct from an architecture-selection failure: the correct arm64 binary launches, the command succeeds, and only the caught diagnostic leaks.

MrLesk:
Fixes #803.

The npm launcher's Rosetta probe ran `execFileSync("/usr/sbin/sysctl", ...)` without a `stdio` option, so in restricted macOS shells the child's `sysctl: Operation not permitted` leaked to stderr before every otherwise-successful command, breaking output parsing for agents and CI. The exception itself was already caught; only the diagnostic leaked.

The probe now passes `stdio: ["ignore", "pipe", "ignore"]` (stdout stays piped for the return value) and the exec function is injectable, mirroring the file's existing `resolveBinaryPath(..., resolver)` convention, so tests pin the exact options without mutating the child_process builtin.

Verification: end-to-end stderr measurement with real child processes (0 bytes with the fix vs 32 bytes before, identical return values on success, denial, and non-darwin paths), five new regression tests confirmed to fail without the fix, `node --check` on both shipped CJS files, and a green full suite. Independently reviewed with no blocking findings.

### DIFF
--- a/backlog/tasks/back-578 - Stop-the-Rosetta-sysctl-probe-from-leaking-stderr.md
+++ b/backlog/tasks/back-578 - Stop-the-Rosetta-sysctl-probe-from-leaking-stderr.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-578
 title: Stop the Rosetta sysctl probe from leaking stderr
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@claude'
 created_date: '2026-08-07 17:25'
+updated_date: '2026-08-07 18:20'
 labels:
   - bug
 dependencies: []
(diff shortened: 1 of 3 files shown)


MrLesk:
@codex review

PR #860, opened 2026-08-07
BACK-575 - Fail fast instead of silently losing concurrent task edits

### LINKED ISSUE #843: [Bug]: `task edit` loses concurrent writes silently — #565's lock was never applied to the edit path

MrLesk:
Fixes #843.

`updateTaskFromInput` was an unlocked read-modify-write serving CLI `task edit`, MCP `task_edit`, and the web PUT: two concurrent edits of the same task silently lost one write while both reported success (measured: 7 of 8 concurrent writes lost pre-fix).

The edit funnel now runs inside a filesystem-level task lock (proper-lockfile, `retries: 0`) with an in-lock re-read. On contention the second writer fails fast with `Edit failed: TASK-X is being modified by another process; retry if appropriate.` — no waiting, no merging, no auto-retry; the caller decides. CLI exits non-zero, the web server returns 409, MCP returns OPERATION_FAILED. The Draft-demotion branch is locked at `demoteTaskWithUpdates` itself so both its entry points (web funnel and MCP's direct call) are covered. Locks live under `backlog/.locks/` (per-checkout, so sibling worktrees never falsely contend); a crashed holder's stale lock auto-breaks after 10s.

Design reference and test-harness shape come from @iRonin's withdrawn PR #852 and the excellent #843 report; semantics differ deliberately (fail-fast per maintainer decision, vs wait-and-re-read).

Verification: 9-test concurrency suite plus a parallel smoke script (8 concurrent real-CLI writers: exactly one winner, losers loud, file content matches the winners exactly); each test confirmed to fail on the unfixed code; kill -9 stale-lock recovery verified; independently reviewed including a live exploit of the demote race, which is now closed. Full suite 1906 pass / 0 fail on the rebased head.

### DIFF
--- a/backlog/tasks/back-575 - Fail-fast-instead-of-silently-losing-concurrent-task-edits.md
+++ b/backlog/tasks/back-575 - Fail-fast-instead-of-silently-losing-concurrent-task-edits.md
@@ -5,7 +5,7 @@ status: In Progress
 assignee:
   - '@claude'
 created_date: '2026-08-07 17:25'
-updated_date: '2026-08-07 17:53'
+updated_date: '2026-08-07 18:39'
 labels:
   - bug
 dependencies: []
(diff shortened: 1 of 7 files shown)


MrLesk:
Alex's agent: We verified this scenario empirically with three schedules: the reported interleaving, strict sequential A-then-B, and the proposed lock-before-read ordering all end in the identical state (`labels: []`), so the proposed change does not alter the outcome of its own reproduction. B's `labels: []` is an explicit replacement applied to the task as re-read inside the lock — nothing from the stale snapshot is written; the result is ordinary last-writer-wins, the same state as if B's request had arrived a moment later with no concurrency at all. (Acquiring the lock "before any lookup" also isn't possible as stated: the lock targets the task file, and resolving the ID to a path is a lookup.)

The lock's contract is fail-fast on overlapping writers plus an in-lock re-read, which guarantees every final state is explainable by a sequential ordering of the operations that reported success, with overlapping losers failing loudly. What remains is a client deciding `labels: []` from a stale view and then writing it — that lives at the caller and would need optimistic versioning (e.g. an If-Match/updatedDate precondition on the web PUT), a separate feature rather than a lock-placement change, noted as a follow-up candidate.

PR #871, opened 2026-08-07
BACK-582 - Add a decision list command

### LINKED ISSUE #845: [Feature]: `decision` has no `list` — decisions are writable and searchable but not enumerable
**CLI-only counterpart to the Web list-view request in #778.** This asks for a `backlog decision list`
subcommand with filtering and machine-readable output — it does *not* ask for another Web page.

## Summary

`backlog decision` exposes exactly one subcommand:

```
$ backlog decision --help
Commands:
  create [options] <title>
  help [command]
```

So there is no way to answer *"which decisions are in status X?"* — the question a decision log
exists to answer. `task` has `list` with `--status`, `--priority`, `--labels`, `--plain`, `--json`;
`decision` has none of it.

`backlog search` does index decisions, but it is fuzzy and scored, not a filter — searching
`proposed` returns every decision with near-identical low scores whether or not it has that status.
The only reliable listing today is:

```sh
grep -l "^status: proposed" backlog/decisions/*.md
```

## Why this is worth having

`decision create -s <status>` accepts **arbitrary** status values (verified — `ratified-by-user` is
accepted, as is a value containing a slash and an emoji). That is genuinely useful: it lets a team
distinguish a proposal an AI agent wrote from a ruling a human actually made, which otherwise read
identically in prose.

But the convention only pays off if the states are enumerable. Being able to run
`backlog decision list --status proposed --plain` is what turns it from a filing convention into a
queue someone can drain.

## Suggested shape

```
backlog decision list [-s <status>] [--plain] [--json]
```

Mirroring `task list` output would be enough. Two smaller adjacent gaps, mentioned only for
context — happy to split them out if you'd prefer separate issues:

- `decision create` accepts only `-s`, so the `## Context` / `## Decision` / `## Consequences`
  bodies must be written by hand afterwards. A `-d/--description` or `--context` would let an agent
  create a complete record in one call.
- `decision create` rejects `--plain` (`error: unknown option '--plain'`), which is surprising when
  the documented guidance for AI agents is "always use `--plain`".

Immutability is correct and I'm not asking for `decision update` — superseding with a new record is
proper ADR practice.

## Environment

backlog.md 1.49.3 (homebrew) · macOS arm64.

MrLesk:
Fixes #845.

Decisions were writable and searchable but not enumerable: `backlog decision` exposed only `create`, and `decision create --plain` errored even though shipped agent guidance says to always pass `--plain`. `backlog decision list` now works with `--plain` and `--json` — rows show id, title, and status as stored (statuses are free-form on decisions), the JSON envelope follows the versioned shape `task list` uses and reuses the exact decision summary object `search --json` emits, and `--json --plain` conflicts identically to the other read commands via the shared output-mode resolver. `decision create` also now accepts `--plain` (accept-and-proceed; creation already prints one plain line and no decision detail format exists yet) and gained the help schema it never had.

Deliberately out of scope per the task: MCP tools, status filters, interactive picker (there is no `decision view` to open), and identity/duplicate handling (owned by BACK-580, verified non-overlapping).

Verification: all modes exercised empirically at 0/1/many decisions including free-form statuses and duplicate-ID files, JSON shape confirmed identical between surfaces, non-vacuous tests, full suite 1972 pass / 0 fail. Independently reviewed with no blocking findings.

### DIFF
--- a/backlog/tasks/back-582 - Add-a-decision-list-command.md
+++ b/backlog/tasks/back-582 - Add-a-decision-list-command.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-582
 title: Add a decision list command
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@claude'
 created_date: '2026-08-07 17:25'
+updated_date: '2026-08-07 22:51'
 labels:
   - enhancement
 dependencies: []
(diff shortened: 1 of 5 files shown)


MrLesk:
@codex review

PR #875, opened 2026-08-08
BACK-608 - Remove gray-matter cache poisoning with a shared no-cache parse wrapper

MrLesk:
## Problem

`gray-matter` keeps a module-level cache of parse results keyed by the input string and hands the **same** `data` object back to every later caller of identical content. Two consequences bit us twice independently during the Aug 2026 review rounds:

- any caller that mutates its parse result silently changes what the next parse of that content returns;
- a malformed document only throws on its **first** parse, so later parses of the same bytes look like a file with empty frontmatter.

Both incidents were patched locally rather than structurally: `parseMarkdown` passed an options object to dodge the cache, and `parseAssigneeConfigValue` switched to `Bun.YAML.parse`. Every other call site was still exposed.

## Fix

One shared wrapper, `src/markdown/frontmatter.ts`, is now the only module that imports `gray-matter`:

- `parseFrontmatter(content)` -> `matter(content, {})`, returning `{ data, content }`
- `stringifyFrontmatter(content, data)` -> `matter.stringify(content, data, {})`

`gray-matter` 4.0.3 only reads and writes its cache when the `options` argument is falsy (`node_modules/gray-matter/index.js`: *"only cache if there are no options passed"*), so passing an options object opts out of both. Every call site now routes through the wrapper:

- `src/markdown/parser.ts` — `parseMarkdown`
- `src/markdown/serializer.ts` — `serializeTask` / `serializeDecision` / `serializeDocument`
- `src/file-system/operations.ts` — `parseConfigListValues`, `parseDefinitionOfDoneFromYaml`
- `src/core/backlog.ts` — `updateDecisionFromContent` (was a dynamic `import("gray-matter")`)

`parseAssigneeConfigValue` deliberately uses `Bun.YAML.parse` for a single scalar/list value and stays as it is. No frontmatter format, engine, option, or serializer-output change: parsing semantics per site are unchanged, and the serializer keeps writing byte-identical output. Two stale test comments telling readers that fixtures need distinct bodies to dodge the parse cache were dropped.

## Evidence

- Mechanism verified against the installed version, not assumed: with no options, a second parse of the same string returned the mutated title and the malformed fixture threw only on the first attempt; with `{}`, `matter.cache` stayed empty, the second parse was pristine, and the malformed fixture threw every time. A shared options object behaves the same as a fresh one.
- New regression tests in `src/test/markdown.test.ts` (`frontmatter parse cache`): one mutates `parseFrontmatter`'s returned `data` (overwrite, extra key, array push) and `content`, then asserts the next parse of the identical string is pristine; one mutates a `parseMarkdown` result and asserts `parseTask` on the same string still sees the real `id`/`title`/`labels`.
- Those tests were confirmed to catch the old behavior: temporarily reverting the wrapper to a bare `matter(content)` failed both new tests (`title: "mutated title"`, injected key, `id: ""`) plus the pre-existing malformed-frontmatter test; restoring the fix turns them green.
- Serializer parity: `matter.stringify(body, data)` and `matter.stringify(body, {...data}, {})` produce byte-identical strings, and the serializer suites pass.
- `bunx tsc --noEmit` clean, `bun run check .` clean (367 files), `bun run test` green — 2062 pass, 6 skip, 0 fail across 222 files.

Task: BACK-608

### DIFF
--- a/backlog/tasks/back-608 - Remove-gray-matter-cache-poisoning-with-a-shared-no-cache-parse-wrapper.md
+++ b/backlog/tasks/back-608 - Remove-gray-matter-cache-poisoning-with-a-shared-no-cache-parse-wrapper.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-608
 title: Remove gray-matter cache poisoning with a shared no-cache parse wrapper
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@Claude'
 created_date: '2026-08-08 15:56'
+updated_date: '2026-08-08 16:22'
 labels: []
 dependencies: []
 ordinal: 247000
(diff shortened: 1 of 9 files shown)


MrLesk:
@codex review

Issue #854, opened 2026-08-07
[Bug]: init ignores BACKLOG_CWD and can reinitialize a different existing project


iRonin:
**Describe the bug**

`BACKLOG_CWD` pins normal CLI commands to a project root, but `backlog init` ignores it and always uses the process working directory. If the process directory already contains a Backlog.md project while `BACKLOG_CWD` points somewhere else, `init` re-initializes the process-directory project, exits 0, and leaves the requested target untouched.

The command reports that it found an existing project, but never says that `BACKLOG_CWD` was ignored or prints the absolute project it is about to rewrite.

**To Reproduce**

The control and reproduction below use two scratch projects so no real board is involved:

```sh
B=/opt/homebrew/bin/backlog
ROOT=$(mktemp -d)
mkdir "$ROOT/caller" "$ROOT/target"

(
  cd "$ROOT/caller"
  "$B" init "Caller sentinel" --defaults --no-git \
    --integration-mode none --default-editor true \
    --auto-open-browser false --task-prefix call
)
(
  cd "$ROOT/target"
  "$B" init "Target control" --defaults --no-git \
    --integration-mode none --default-editor true \
    --auto-open-browser false --task-prefix tgt
)

# Control: BACKLOG_CWD is valid and normal CLI root resolution honors it.
(
  cd "$ROOT/caller"
  BACKLOG_CWD="$ROOT/target" "$B" config get projectName
)
# Target control

shasum -a 256 \
  "$ROOT/caller/backlog/config.yml" \
  "$ROOT/target/backlog/config.yml"
# 503b434c3417dcc0bf30a24db6c33a4ac6e1c8138f7093ed468cd3f4c66a981c  caller/backlog/config.yml
# a08cb13f03f02c6edb40d6b3664dd204a51e7ca6fed343faa9d7e9380e803e3b  target/backlog/config.yml

# Reproduction: the same pin is ignored by init.
(
  cd "$ROOT/caller"
  BACKLOG_CWD="$ROOT/target" "$B" init "Override requested" \
    --defaults --no-git --integration-mode none
)
# Existing backlog project detected. Current configuration will be preserved where not specified.
# ...
# Updated backlog project configuration: Override requested

grep '^project_name:' \
  "$ROOT/caller/backlog/config.yml" \
  "$ROOT/target/backlog/config.yml"
# caller/backlog/config.yml:project_name: "Override requested"
# target/backlog/config.yml:project_name: "Target control"

shasum -a 256 \
[...]

PR #876, opened 2026-08-08
BACK-605 - Route all TUI operations through shared core with runtime cwd

MrLesk:
## Problem

The board TUI built its own `Core` from `process.cwd()` in every mutation handler:

```
src/ui/board.ts:1025  new Core(process.cwd(), { enableWatchers: true })   # create task
src/ui/board.ts:1215  ...                                                # edit task
src/ui/board.ts:1309  ...                                                # complete (details popup)
src/ui/board.ts:1347  ...                                                # archive (details popup)
src/ui/board.ts:1396  ...                                                # reorder / move
src/ui/board.ts:1554  ...                                                # complete (board key)
src/ui/board.ts:1597  ...                                                # archive (board key)
```

That bypasses the runtime-cwd resolution every CLI entry point goes through
(`resolveRuntimeCwd`/`requireRuntimeCwd`, added for #854 in BACK-581), so `BACKLOG_CWD`
was silently ignored by the board's own operations. It also diverged from the project root
the board *renders* from (`requireProjectRoot()`), which means the same bug appeared
whenever the TUI was launched from a subdirectory: the board showed one project and `n`/`e`/`c`/`a`/`m`
wrote to another. `src/ui/enhanced-views.ts` and the `src/ui/task-viewer-with-search.ts`
fallback had the same construction, and three non-TUI call sites resolved their own cwd too.

Direction from Alex: all interfaces go via core and read the same core logic — one
resolution path, no per-interface divergence.

## Fix

Pure plumbing; no TUI feature or UX changes.

- **One resolution path.** New `createRuntimeCore(options?)` exported next to `Core` in
  `src/core/backlog.ts` builds a `Core` from `resolveRuntimeCwd()`, so `--cwd`/`BACKLOG_CWD`
  (else `process.cwd()`) is honoured in exactly one place, and an invalid override fails
  closed like the CLI does.
- **Board reuses the caller's instance.** `renderBoardTui` accepts `core?: Core`; all seven
  handlers now go through a single `getCore()` closure that returns the supplied instance.
  A memoised `createRuntimeCore({ enableWatchers: true })` fallback covers only direct
  callers with no instance (the existing TUI tests). `unified-view.ts`,
  `simple-unified-view.ts` and `enhanced-views.ts` pass the `Core` they already hold, so
  production constructs fewer Cores than before, not more.
- **Remaining TUI sites.** `enhanced-views.ts` uses the threaded instance;
  `task-viewer-with-search.ts`'s optional-core fallback uses `createRuntimeCore`.

## Interface audit

Aligned on the shared factory:

| Site | Was |
| --- | --- |
| `src/utils/task-path.ts` `getTaskPath` fallback | `new Core(process.cwd())` |
| `src/utils/status.ts` `getValidStatuses` fallback | `new Core(process.cwd())` |
| `src/completions/data-providers.ts` | local `createCore()` on `process.cwd()` |

Already resolver-based, unchanged:

- **Web server** — `new BacklogServer(cwd)` (`src/cli.ts:5230`) with `cwd` from
  `requireProjectRoot()` → `requireRuntimeCwd()` → `resolveRuntimeCwd()`.
- **MCP** — `createMcpServer(runtimeCwd.cwd)` from `src/commands/mcp.ts`, which calls
  `resolveRuntimeCwd({ cwd: options.cwd })`.
- Every `new Core(cwd)` in `src/cli.ts` takes `cwd` from `requireProjectRoot()`/`requireRuntimeCwd()`.

`process.cwd()` uses deliberately left alone, because they are not project resolution:

- `src/server/index.ts:483` — saves/restores the process cwd around the bundled-asset `chdir`; builds no Core.
- `src/readme.ts` — writes `README.md` and a temp board file relative to `process.cwd()` (output path).
- `src/commands/help-schema.ts:183` — output-directory override for schema generation.

## Evidence

**No Core is built from `process.cwd()` any more:**

```
$ grep -rn "Core(process.cwd" src/
$ echo $?
1
```

**Regression tests.** `src/test/tui-runtime-cwd.test.ts` drives the board's `n` handler
through the handler's own `persist` with `process.cwd()` pointed at an unrelated directory:
one case asserts the task lands in `BACKLOG_CWD`, the other asserts a supplied `Core`
instance wins over `BACKLOG_CWD`. Both fail against the pre-change `board.ts`:

```
$ git checkout src/ui/board.ts && bun test --timeout=20000 src/test/tui-runtime-cwd.test.ts
error: expect(received).toHaveLength(expected)
Expected length: 1
Received length: 0
(fail) board TUI runtime working directory > writes into BACKLOG_CWD when no Core instance is supplied
(fail) board TUI runtime working directory > reuses the supplied Core instance instead of resolving again
 0 pass
 2 fail
```

and pass with the fix:

```
$ bun test --timeout=20000 src/test/tui-runtime-cwd.test.ts
 2 pass
 0 fail
```

`src/test/runtime-cwd.test.ts` adds `createRuntimeCore` cases for the `process.cwd()`
default, the `BACKLOG_CWD` override, and rejection of an invalid override.

**End-to-end PTY run of a real mutation.** cwd = this repo worktree (a different Backlog
project, `BACK-` prefix); `BACKLOG_CWD` = a `/tmp` fixture project holding one task.

```
$ BACKLOG_CWD=$FIXTURE bun src/cli.ts board          # rendered from an unrelated cwd
Project: Project
| To Do | In Progress | Done |
| **TASK-1** - Fixture only task |  |  |

$ BACKLOG_CWD=$FIXTURE expect -f board-archive.expect   # spawn board, send "a", Enter, "q"
ARCHIVED footer seen
$ find $FIXTURE/backlog -name "*.md"
$FIXTURE/backlog/archive/tasks/task-1 - Second-fixture-task.md
```

The same script against the pre-change `board.ts` timed out waiting for the footer and left
the file in `backlog/tasks/` (the board resolved the repo's project, where `TASK-1` does not exist):

```
TIMEOUT waiting for archived footer
$ ls $FIXTURE/backlog/tasks
task-1 - Second-fixture-task.md
```

**Checks:**

```
$ bunx tsc --noEmit      # clean
$ bun run check .        # Checked 367 files. No fixes applied.
$ bun run test           # 2065 pass, 6 skip, 0 fail (2071 tests, 223 files)
```

Behaviour is unchanged when `BACKLOG_CWD` is unset: `resolveRuntimeCwd()` returns
`process.cwd()`, and in the CLI-driven TUI the board now shares the very `Core` the task
viewer was already using.

Task record: `BACK-605` (plan, notes, checked acceptance criteria and Definition of Done are in this branch).

### DIFF
--- a/backlog/tasks/back-605 - Route-all-TUI-operations-through-shared-core-with-runtime-cwd.md
+++ b/backlog/tasks/back-605 - Route-all-TUI-operations-through-shared-core-with-runtime-cwd.md
@@ -1,11 +1,11 @@
 ---
 id: BACK-605
 title: Route all TUI operations through shared core with runtime cwd
-status: In Progress
+status: Done
 assignee:
   - '@Claude'
 created_date: '2026-08-08 15:56'
-updated_date: '2026-08-08 16:04'
+updated_date: '2026-08-08 16:23'
 labels: []
 dependencies: []
 ordinal: 244000
@@ -19,17 +19,17 @@ TUI board handlers construct Core(process.cwd()) directly at multiple sites in s
 
 ## Acceptance Criteria
 <!-- AC:BEGIN -->
-- [ ] #1 No TUI code constructs Core from process.cwd() directly; all handlers reuse the shared core instance or the shared runtime cwd resolver
-- [ ] #2 TUI operations honor BACKLOG_CWD end to end
-- [ ] #3 A repo-wide audit confirms no other interface constructs Core from process.cwd() outside the shared resolver, or aligns the stragglers
-- [ ] #4 Tests or recorded verification evidence cover the BACKLOG_CWD path in the TUI
+- [x] #1 No TUI code constructs Core from process.cwd() directly; all handlers reuse the shared core instance or the shared runtime cwd resolver
+- [x] #2 TUI operations honor BACKLOG_CWD end to end
+- [x] #3 A repo-wide audit confirms no other interface constructs Core from process.cwd() outside the shared resolver, or aligns the stragglers
+- [x] #4 Tests or recorded verification evidence cover the BACKLOG_CWD path in the TUI
 <!-- AC:END -->
 
 ## Definition of Done
 <!-- DOD:BEGIN -->
-- [ ] #1 bunx tsc --noEmit passes when TypeScript touched
-- [ ] #2 bun run check . passes when formatting/linting touched
-- [ ] #3 bun test (or scoped test) passes
+- [x] #1 bunx tsc --noEmit passes when TypeScript touched
+- [x] #2 bun run check . passes when formatting/linting touched
+- [x] #3 bun test (or scoped test) passes
 <!-- DOD:END -->
 
 ## Implementation Plan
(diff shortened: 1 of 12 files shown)


MrLesk:
Addressed the P2 on `src/core/backlog.ts` in 7ce0b691.

**The finding was right.** `createRuntimeCore` used `resolveRuntimeCwd()` verbatim, but the CLI goes `requireProjectRoot()` → `findBacklogRoot()`, which *ascends* from the resolved directory to the real project root. So with `BACKLOG_CWD` (or the shell cwd, for fallback callers) pointing at a subdirectory of a valid project, the factory targeted `<subdir>/backlog`, which does not exist — completions silently returned empty lists while CLI commands on the same environment worked. That is exactly the cross-interface divergence this PR exists to remove, so it is fixed here rather than deferred.

**Fix:**

```ts
export async function createRuntimeCore(options?: { enableWatchers?: boolean }): Promise<Core> {
	const { cwd } = await resolveRuntimeCwd();
	return new Core((await findBacklogRoot(cwd)) ?? cwd, options);
}
```

**Not-found semantics.** When `findBacklogRoot` returns null the factory keeps the resolved directory as the root, rather than exiting the way `requireProjectRoot` does. Every caller is an optional-core fallback in library code, not a CLI entry point, so this preserves each one's pre-existing behaviour: completions degrade to their static fallbacks, `getTaskPath` returns null, and nothing kills the user's shell. An invalid `BACKLOG_CWD` still rejects out of `resolveRuntimeCwd` (fail closed), which `withCore` swallows into the static fallback. No semantics fork here — `process.exit(1)` was never a viable option for these call sites.

**Coverage** (`src/test/runtime-cwd.test.ts`, +6 tests): `BACKLOG_CWD` at a nested subdirectory resolves to the parent project root; same via `process.cwd()`; no-project stays on the resolved directory; and the completion providers (`getTaskIds`/`getLabels`/`getAssignees`) read the parent project from a subdirectory, plus two graceful-degradation cases. The three subdirectory assertions fail without the ascent, reproducing the empty-`getTaskIds()` repro:

```
(fail) createRuntimeCore > ascends to the project root when BACKLOG_CWD points at a subdirectory
Expected: ".../project"
Received: ".../project/packages/web/src"
(fail) createRuntimeCore > ascends to the project root from a subdirectory of process.cwd()
(fail) shell completion data providers > reads the parent project when BACKLOG_CWD points at a subdirectory
 11 pass  3 fail
```

**CLI-level confirmation** with `BACKLOG_CWD=<fixture>/packages/cli` inside an initialized fixture project:

```
# after
$ bun src/cli.ts completion __complete "backlog task view " 18
TASK-1
--plain
--json

# before
$ bun src/cli.ts completion __complete "backlog task view " 18
--plain
--json
```

**Checks:** `bunx tsc --noEmit` clean, `bun run check .` clean (367 files), `bun run test` → 2071 pass, 6 skip, 0 fail (2077 tests, 223 files).

Leaving the review thread open for you to resolve.

PR #895, opened 2026-08-09
BACK-651 - Remove the temp-file roundtrip in README board export

pxmpsdev:
Removes `.temp-board.md` roundtrip in `src/readme.ts` (race, stale file, Bun.$ shell) and makes `src/board.ts:100` non-mutating (`[...items].sort()`).\n\nVerified: `bunx tsc --noEmit` ok, `bun run check` 368 files ok, board tests 11 pass.

PR #896, opened 2026-08-09
fix: avoid mutating subtask array when sorting in board generation

pxmpsdev:
Consistent with prior `items` fix: `src/board.ts:131` mutates `children` map via `subs.sort()`. Use `[...subs].sort()` with `sortedSubs` to keep render pure.\n\nVerified: `bunx tsc --noEmit` ok, `bun run check` 369 files ok, board tests 8 pass. Small focused fix.

### DIFF
--- a/src/board.ts
+++ b/src/board.ts
@@ -129,12 +129,12 @@ Project: ${projectName}
 		for (const t of top) {
 			result.push(t);
 			const subs = children.get(t.id) || [];
-			subs.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


MrLesk:
Thanks for the PR. The array sorted here is function-local: it's built a few lines above in the same function and never escapes it, so the in-place sort cannot affect any caller's data and the defensive copy doesn't change behavior. Closing since there's no bug to fix — see also the sibling change in #895, where the README export cleanup from that PR is being taken forward.

— Alex's agent

PR #897, opened 2026-08-09
BACK-648 - Fall back to a placeholder filename for punctuation-only titles

pxmpsdev:
Previously `sanitizeFilename('!!!')` → `''` gave filenames like `task-42 - .md`. Fallback to id-only (`task-42.md`, `decision-1.md`, `doc-1.md`) when sanitized title is empty. Covers tasks, decisions, documents in `src/file-system/operations.ts:661,1126,1162`.\n\nVerified: `bunx tsc --noEmit` ok, `bun run check` 369 files ok.


MrLesk:
@codex review

PR #904, opened 2026-08-10
BACK-588 - Make the TUI help popup robust to resize and wrapped lines

MrLesk:
## Summary

- reflow an open help popup when terminal dimensions change
- derive scroll bounds from rendered wrapped rows and clamp offsets after resize
- remove the resize listener on close and keep footer hints in sync

## Testing

- 27 focused TUI/help tests passed
- bunx tsc --noEmit
- bun run check .
- bun run build
- git diff --check
- real-screen coverage for 24-to-10 row resize and 30x24 wrapped shortcuts


MrLesk:
@codex review

PR #909, opened 2026-08-10
BACK-590 - Support mouse clicks in the TUI task composer

MrLesk:
## Summary

- route text and selector clicks through the existing composer focus transition
- enter text read mode with a visible caret and exclusive focus styling
- open all three selector pickers from pointer activation

## Testing

- 31 focused composer tests passed
- deterministic real-widget click, caret, focus, text-entry, and picker evidence
- bunx tsc --noEmit
- bun run check .
- bun run build
- git diff --check


MrLesk:
@codex review

PR #912, opened 2026-08-10
BACK-626 - Make task lifecycle commands local-first

shixi-li:
## Summary

- make `task archive`, `task complete`, and `task demote` resolve their targets from the local working-copy task corpus
- pass the local-only option through the mutation boundary while preserving the existing default behavior for non-CLI Core callers
- align branch-only misses with the same actionable hint used by `task view` and `task edit`
- preserve fail-closed `AmbiguousTaskIdError` behavior for active/completed ID collisions

## Why

These three CLI commands still used the default cross-branch lookup twice: once during command preflight and again inside the Core mutation. That could trigger a remote fetch/full corpus load and produced different errors for tasks that exist only on another branch, despite the owner ruling that CLI task commands are local-only.

The Core methods now accept an optional third `TaskReadOptions` argument. Existing callers retain their current defaults; only the three CLI handlers opt into `includeCrossBranch: false` for both lookup stages.

## Validation

- `bun test --timeout=10000 src/test/local-task-command-performance.test.ts` — 9 passed
- lifecycle/Core/MCP scoped suite — 142 passed
- `bunx tsc --noEmit`
- `bun run check .`
- `bun run build`
- independent diff review: P0=0, P1=0

Closes BACK-626.

### DIFF
--- a/backlog/tasks/back-626 - Make-task-archive-complete-and-demote-local-first-like-view-and-edit.md
+++ b/backlog/tasks/back-626 - Make-task-archive-complete-and-demote-local-first-like-view-and-edit.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-626
 title: 'Make task archive, complete, and demote local-first like view and edit'
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@shixi-li'
 created_date: '2026-08-10 06:10'
+updated_date: '2026-08-10 10:45'
 labels: []
 dependencies: []
 ordinal: 262000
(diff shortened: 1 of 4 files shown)


shixi-li:
Hi maintainers — the fork [CI workflow](https://github.com/MrLesk/Backlog.md/actions/runs/31380679724) for this PR is currently awaiting approval. The local validation results are included in the PR description. When convenient, could you approve the workflow so the change can be evaluated? Thanks!

MrLesk:
@codex review

PR #607, opened 2026-04-25
BACK-412 - Add touched-files field to tasks and filename-based search

MrLesk:
## Summary

Adds a dedicated `modifiedFiles` task field, persisted as `modified_files` frontmatter, so tasks can record project-root-relative file paths they touched.

Wires the field through task parsing/serialization, Core create/edit flows, CLI create/edit/search, server search, MCP create/edit/search schemas, shared search indexes, plain text output, and TUI detail display. MCP `task_search` now supports a separate `modifiedFiles` filter so file-path filtering does not have to share the generic query string.

Updates shipped agent/MCP guidance and README notes for setting and searching modified files.

## Validation

- `bun test`
- `bunx tsc --noEmit`
- `bunx biome check README.md src/cli.ts src/core/backlog.ts src/core/search-service.ts src/formatters/task-plain-text.ts src/guidelines/agent-guidelines.md src/guidelines/mcp/overview-tools.md src/guidelines/mcp/overview.md src/markdown/parser.ts src/markdown/serializer.ts src/mcp/tools/tasks/handlers.ts src/mcp/tools/tasks/index.ts src/mcp/tools/tasks/schemas.ts src/mcp/utils/schema-generators.ts src/server/index.ts src/test/cli-refs-docs.test.ts src/test/cli-search-command.test.ts src/test/markdown.test.ts src/test/mcp-tasks.test.ts src/test/search-service.test.ts src/test/server-search-endpoint.test.ts src/test/task-search-label-filter.test.ts src/types/index.ts src/types/task-edit-args.ts src/ui/task-viewer-with-search.ts src/utils/modified-files.ts src/utils/task-edit-builder.ts src/utils/task-search.ts src/web/lib/api.ts`

## Notes

`bun run check .` is still blocked by an unrelated pre-existing `package.json` formatting issue; the staged source/docs files pass Biome.

PR #913, opened 2026-08-10
BACK-633 - Show and edit modified files in the web task modal

ivan812205:
## Summary

The `modifiedFiles` field was added in #607 (BACK-412) and is wired through the whole stack: the markdown parser/serializer, `backlog task edit --modified-files` in the CLI, the plain output, the TUI task viewer, the MCP tools, and web search. The one surface that never caught up is the web task details modal: the card neither displayed the files nor offered a way to edit them.

The asymmetry is easy to reproduce today: in the web UI you can *search* tasks by file name — `search-command-query.ts` parses `modifiedfiles:` tokens, `SideNavigation.tsx` detects them as a command query, and `lib/api.ts` forwards them as `modifiedFile` params — but opening a task found that way showed no trace of the files that matched it.

This PR adds a **Modified files** section to `TaskDetailsModal.tsx`, modeled 1:1 on the existing References section right above it:

- the file paths render as a monospace list (`<code>`, `break-all` for long paths);
- each row gets a remove button revealed on hover (hidden for read-only cross-branch tasks);
- an add form (preview mode only) appends a trimmed, de-duplicated path;
- both mutations go through the existing `handleInlineMetaUpdate`, i.e. the same optimistic-update path References use.

No server changes were needed: `PUT /api/tasks/:id` already accepts `modifiedFiles` (`src/server/index.ts`), so the backend was ready and waiting. The new state is wired into the modal's form-state snapshot, the dirty-preserving refresh logic, task-switch reset, and external-update sync, same as the other list fields.

One file changed: `src/web/components/TaskDetailsModal.tsx` (+76).

## Validation

- `bun test --timeout=10000` on this branch: **2244 pass, 6 skip, 1 fail** (2251 tests across 230 files).
- The single failure is `Git Operations > fetch > hard-kills a timed-out Git process group even when a child holds its pipes` — a pre-existing timing flake unrelated to this change: with this diff stashed, the same test fails on unmodified `main` as well, and it passes 5/5 consecutive isolated runs with the diff applied (failures land at ~504–508 ms against its 500 ms kill window).
- `bunx tsc --noEmit`: clean.

## Notes

- On one full-suite run `BacklogServer statistics endpoint > reconciles a selected backlog root before reading its statistics` also failed once under load; it passes in isolation both with and without this diff, so it looks like the same class of environment-sensitive flake.
- This UI-only change follows the pattern of the section directly above it, so no new tests were added; the search/API/server behavior it exposes is already covered by existing suites.
- There is no backlog task for this yet — happy to add one (`backlog task create`) or adjust scope if you'd prefer this discussed in an issue first.

---

## Maintainer follow-up (BACK-633)

Thanks @ivan812205 — the diagnosis and the diff were both right, so this was picked up and finished in place rather than reworked. Your commit is the base of the branch; the commits on top are maintainer edits.

Adding the backlog task: **BACK-633 - Show and edit modified files in the web task modal**, created with the CLI and committed to this branch.

### Review of the original change

Verified rather than assumed:

- `PUT /api/tasks/:id` does apply the field (`src/server/index.ts`), and core normalizes it (`resolveModifiedFiles` in `src/core/backlog.ts`), so no server change was needed — correct as described.
- The refresh-preservation, task-switch reset, and external-update sync all match the sibling `references` handling.
- `handleSave` deliberately omits `modifiedFiles`, exactly as it omits `references`, so a Save in edit mode cannot clobber an inline edit. Correct as written.
- Branch was already on top of current `main`, so no rebase was needed.

### Added: the section has to survive a large task

A finished task can list well over a hundred paths, and paths can be long enough to wrap several times. Unbounded, the section pushed Documentation, Acceptance Criteria, Plan, Notes and Comments out of reach.

- The paths list is capped (`max-h-64`, `overflow-y-auto`, `overscroll-contain`), following the bounded-list pattern already used by `CleanupModal` and `DuplicateIdRepairModal`. `overscroll-contain` keeps the inner wheel from scrolling the modal behind it.
- The heading carries the count — `Modified files (130)` — matching the existing `Comments (n)` precedent, so a capped list still announces its size without adding helper text.
- Rows use `items-start` with a small offset on the remove button, so on a path that wraps to three lines the control sits with the first line instead of floating in the middle of the row.
- Long paths keep your `break-all` wrapping rather than truncating, so no path is hidden.
- `hasCreateModeEntries` now counts `modifiedFiles` alongside `references`.

Measured in a throwaway project on a task seeded with 130 long paths (Chrome, 1280px):

| | unbounded | capped |
|---|---|---|
| Section height | 10508px | **372px** |
| Modal scroll height | 11475px | **1339px** |

All 130 paths stay in the DOM and the last is reachable by scrolling (list content 10392px inside a 256px box). No horizontal overflow at 1280px or at 375px, where paths wrap to six lines and the section still measures 372px. The cap is inert for ordinary tasks: a 3-path task measures an 88px list with no scrollbar.

### Added: tests

`src/test/web-task-details-modal-modified-files.test.tsx` follows `web-task-details-modal-documentation.test.tsx` — five cases covering the populated list, the empty state, the heading count, a 120-path long-path case asserting every path renders inside a bounded scrolling list with later sections still present, and cross-branch read-only hiding both controls. The many-paths case was confirmed to fail when the height cap is removed.

### Validation

- `bunx tsc --noEmit`: clean.
- `bun run check .`: clean.
- `bun run test`: **2250 pass, 6 skip, 0 fail** (231 files). The git-timeout flake you hit did not reproduce.
- `bun run build`: succeeds.
- Live round-trip: adding a path from the modal wrote it into the task markdown and `backlog task view --json` reported 131; removing it from the modal returned both to 130.

One pre-existing behavior noted but deliberately not changed here: pressing Enter in the add input did not submit under synthesized key events. The References add form behaves identically, so it is shipped parity rather than a regression in this PR, and the Add button works in both.

### DIFF
--- a/backlog/tasks/back-633 - Show-and-edit-modified-files-in-the-web-task-modal.md
+++ b/backlog/tasks/back-633 - Show-and-edit-modified-files-in-the-web-task-modal.md
@@ -1,0 +1,84 @@
+---
+id: BACK-633
+title: Show and edit modified files in the web task modal
+status: Done
+assignee:
+  - '@Claude'
+created_date: '2026-08-15 13:12'
+updated_date: '2026-08-15 13:26'
+labels: []
+dependencies: []
+references:
+  - 'https://github.com/MrLesk/Backlog.md/pull/913'
+type: feature
+ordinal: 269000
+---
+
+## Description
+
+<!-- SECTION:DESCRIPTION:BEGIN -->
+The `modifiedFiles` field shipped in BACK-412 (#607) and is wired through the markdown parser/serializer, `backlog task edit --modified-files`, plain and JSON output, the TUI task viewer, the MCP tools, and web search. The web task details modal is the one surface that never adopted it: a task found in the browser by a `modifiedfiles:` search query opened with no trace of the files that matched it, and there was no way to add or remove a path from the browser.
+
+Closing this parity gap means the browser shows the same task model the CLI and TUI already show, and edits go through the existing `PUT /api/tasks/:id` path that already accepts `modifiedFiles`.
+
+The section must stay usable for real-world task sizes. A completed task can list well over a hundred paths, and paths can be long enough to overflow their row, so an unbounded list would push the rest of the modal (documentation, acceptance criteria, plan, notes, comments) out of reach and risk horizontal overflow.
+
+Contributed by @ivan812205 in PR #913.
+<!-- SECTION:DESCRIPTION:END -->
+
+## Acceptance Criteria
+<!-- AC:BEGIN -->
+- [x] #1 The web task details modal shows a Modified files section listing the task's modified file paths
+- [x] #2 A path can be added and removed from the modal, and the change persists through the existing task update API
+- [x] #3 The section stays usable with 100+ paths: its list is height-bounded and scrolls internally instead of pushing later modal sections out of reach
+- [x] #4 Long paths wrap within their row so neither the modal nor the page scrolls horizontally
+- [x] #5 Read-only cross-branch tasks show the paths without add or remove controls
+- [x] #6 A web test covers rendering a task with many long modified file paths
+<!-- AC:END -->
+
+## Definition of Done
+<!-- DOD:BEGIN -->
+- [x] #1 bunx tsc --noEmit passes when TypeScript touched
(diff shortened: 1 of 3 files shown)


MrLesk:
@codex review

PR #926, opened 2026-08-20
BACK-628 - Stop findIdentity rename fallback from publishing freshness without installing the corpus

jafigueroam94:
## Summary

Fixes a pre-existing defect flagged during the PR #899 (BACK-624) review round: `ContentStore.findIdentity`'s rename-fallback (used when a task file is renamed/deleted and no local candidate matches) loaded the full task corpus through `Core`'s publishing loader purely to resolve one task's identity, then discarded everything except that single task.

That publishing loader unconditionally advances `Core.activeBranchFingerprint` to the branch tips observed at load time. Since the fallback never installs the loaded corpus into the `ContentStore`, a later read would see the fingerprint already match the (possibly just-moved) branch tips and skip the full `store.refreshTasks()`, serving stale branch state indefinitely — until an unrelated ref or config change forced a refresh.

Trigger: deleting/renaming-away a task with no local candidate on disk (branch-fallback hydration that *does* find a candidate otherwise forces a healing refresh, so the narrow case is a task with no branch-side copy either).

## Fix

- Threaded an optional `{ publish?: boolean }` through `ContentStore`'s `taskLoader` and `loadTasksWithLoader()`.
- Threaded the same option through `Core.loadContentStoreCorpus`, defaulting to `publish: true` (unchanged behavior for the two legitimate publishing call sites: `loadCurrentContent` and `refreshTasksFromDisk`).
- The `findIdentity` fallback now calls `loadTasksWithLoader(undefined, { publish: false })`, so its throwaway load never advances `activeBranchFingerprint`.
- Extracted the duplicated `waitUntil()` test polling helper (previously only in `content-store.test.ts`) into `test-utils.ts` so it can be reused.

## Test plan

- [x] Added `"keeps serving fresh branch state after a rename fallback that never installed a corpus"` in `core-task-corpus-regressions.test.ts`, modeled on the existing ID-allocation freshness regression test. It reproduces the reviewer's exact repro: warm the corpus, move a branch tip out-of-band via a worktree, delete a local-only task with no branch-side copy (triggering the fallback), then assert a subsequent read observes the moved tip's fresh content.
- [x] Verified with `git stash` that the new test fails pre-fix (observed `"Before ref move"` instead of `"After ref move"`) and passes post-fix.
- [x] `bunx tsc --noEmit` clean.
- [x] `bun run check .` clean on touched files (pre-existing unrelated Biome findings elsewhere in the tree left untouched).
- [x] `bun run test`: 2340 pass / 6 skip / 0 fail across 243 files.

Ref: BACK-628

### DIFF
--- a/backlog/tasks/back-628 - Stop-findIdentity-rename-fallback-from-publishing-freshness-without-installing-the-corpus.md
+++ b/backlog/tasks/back-628 - Stop-findIdentity-rename-fallback-from-publishing-freshness-without-installing-the-corpus.md
@@ -7,7 +7,7 @@ status: Done
 assignee:
   - '@claude'
 created_date: '2026-08-10 06:37'
-updated_date: '2026-08-29 18:21'
+updated_date: '2026-08-29 18:32'
 labels: []
 dependencies: []
 priority: medium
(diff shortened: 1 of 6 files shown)


MrLesk:
@codex review

MrLesk:
@codex review

Alex's agent

MrLesk:
@codex review

PR #927, opened 2026-08-20
BACK-630 - Filter the web dependency picker to locally-resolvable tasks

jafigueroam94:
## Summary
- Follow-up from the BACK-623 (#898) verification round: dependency validation moved to local-only resolution, but the web dependency picker still suggested tasks from the cross-branch search corpus, so picking a branch-only task failed on save with a CLI-oriented error ("...use 'backlog browser'...") that made no sense from inside the browser.
- `DependencyInput` now gets a `suggestableTasks` prop used only for autocomplete filtering (chip display still resolves through the full `availableTasks` corpus so already-saved cross-branch dependency chips keep linking correctly). `TaskDetailsModal` fetches a local-only task list (`apiClient.fetchTasks({ crossBranch: false })`) when it opens and feeds it in, matching exactly what `validateDependencies()` accepts server-side.
- `server/index.ts` now rewrites the missing-dependency error's CLI hint into web-appropriate copy before returning it to the browser (applied to both the create-task and update-task handlers).

## Test plan
- New interactive test (`web-dependency-picker-local-only.test.tsx`): mounts `TaskDetailsModal` with a cross-branch-only task and a local task, mocks `apiClient.fetchTasks` to assert it's called with `crossBranch:false`, and confirms only the local task appears as a suggestion. Verified fail-before/pass-after via `git stash`.
- New server test (`server-dependency-error-copy.test.ts`): exercises the real HTTP API and asserts a missing-dependency error no longer contains "backlog browser". Verified fail-before/pass-after via `git stash`.
- Fixed a regression the change surfaced in an existing test (`web-task-details-modal-unsaved-navigation.test.tsx`) that needed `apiClient.fetchTasks` mocked now that the picker fetches its own suggestion list.
- `bunx tsc --noEmit`: clean.
- `bunx biome check` on touched `.ts` files: clean.
- Full suite: 2341 pass / 6 skip / 1 fail, the 1 fail (`config-commands.test.ts`) is a pre-existing, unrelated flake that reproduces identically on unmodified `main`.

BACK-630

### DIFF
--- a/backlog/tasks/back-630 - Filter-the-web-dependency-picker-to-locally-resolvable-tasks.md
+++ b/backlog/tasks/back-630 - Filter-the-web-dependency-picker-to-locally-resolvable-tasks.md
@@ -5,7 +5,7 @@ status: Done
 assignee:
   - '@claude'
 created_date: '2026-08-10 07:12'
-updated_date: '2026-08-20 22:23'
+updated_date: '2026-08-24 21:38'
 labels: []
 dependencies: []
 priority: medium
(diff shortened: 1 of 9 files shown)


MrLesk:
@codex review

MrLesk:
@codex review

PR #929, opened 2026-08-21
BACK-638 - Allow task list --status to accept several statuses

yss19850810-crypto:
`--status` is the only list filter that does not accept several values. `--exclude-status` (#745) and `--type` both do; `--status` has kept the single-value form it was given in #82 / #112.

## Symptom

```console
$ backlog task list --status "To Do" --plain
TASK-1 - ...        # 4 tasks

$ backlog task list --status "In Progress" --plain
TASK-2 - ...        # 2 tasks

$ backlog task list --status "To Do" --status "In Progress" --plain
No tasks found.     # expected 6
```

Asking for everything unfinished returns nothing. There is no error — an empty list is indistinguishable from genuinely having no such tasks, which is why this is easy to miss.

## Cause

The option has no accumulator, so repeating the flag overwrites the earlier value rather than collecting it. The filter then runs against whichever status came last. It is not an intersection; it is the first value being discarded.

## Change

`--status` now accumulates and splits on commas, exactly as `--exclude-status` does. Both forms work:

```console
backlog task list --status "To Do" --status "In Progress"
backlog task list --status "To Do,In Progress"
```

A single `--status` behaves as before.

## Why three core files

`TaskListFilter.status` widens from `string` to `string | string[]`, mirroring `excludeStatus`. That change made the compiler point out that the same status comparison is implemented separately in three places:

- `src/core/content-store.ts`
- `src/core/backlog.ts`
- `src/file-system/operations.ts`

Changing only the first would have left two paths on the old behaviour. The three copies are left as they are otherwise — deduplicating them is a separate concern and would make this change harder to review.

## Tests

`src/test/cli-status-filtering.test.ts` covers a single status, a repeated flag, comma-separated values, case-insensitivity across several values, and the combination with `--exclude-status`.

Reverting the accumulator makes three of the five fail, so they exercise the fix rather than following it.

## Verification

`bunx tsc --noEmit`, `bun run lint`, and the new plus existing status-filter tests all pass.

`bun run check .` is not included: on Windows the repository's `* text=auto` gives the working tree CRLF endings, which biome rejects wholesale, so the command fails regardless of the change.

### DIFF
--- a/backlog/tasks/back-638 - Allow-task-list-status-to-accept-several-statuses.md
+++ b/backlog/tasks/back-638 - Allow-task-list-status-to-accept-several-statuses.md
@@ -5,7 +5,7 @@ status: Done
 assignee:
   - '@MrLesk'
 created_date: '2026-08-22 12:22'
-updated_date: '2026-08-22 12:50'
+updated_date: '2026-08-23 12:27'
 labels: []
 dependencies: []
 references:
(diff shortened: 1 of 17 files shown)


MrLesk:
@codex review

MrLesk:
@codex review

MrLesk:
@codex review

MrLesk:
Statuses whose configured name contains a comma cannot be matched through this filter currently: comma-separated values are treated as separate statuses by design.

PR #938, opened 2026-08-24
BACK-639 - Add Git-backed live task coordination

### LINKED ISSUE #937: Add Git-backed atomic task coordination
Related Backlog task: BACK-639

Alex's Agent:  ## Problem  Two users or agents can fetch the same task, both assign or start it locally, and begin duplicate work before either Markdown change is pushed. Fetch-before-edit cannot close this distributed race.  ## Proposed behavior  Add an optional Git-backed coordination layer for repositories with remote operations enabled:  - A fixed custom ref per task under `refs/backlog/tasks/*` points to a small JSON blob. - Claim creation, renewal, expired-lease replacement, live-state publication, and release use explicit compare-and-swap via `--force-with-lease`. - `backlog task start` claims first and stops if another owner wins, then updates the Markdown task through the existing core. - Status and assignee may be published as a transient overlay without adding commits to normal branch history. - Markdown remains the durable source of truth; coordination refs contain only transient state. - Projects with remote operations disabled retain their current behavior.  ## CLI surface  - `backlog task start BACK-123` - `backlog task claim BACK-123` - `backlog task claim BACK-123 --renew` - `backlog task release BACK-123` - `backlog task publish BACK-123` - `backlog task claims`  ## Safety requirements  - Exactly one winner for simultaneous claims. - Claims fail closed when the remote is unavailable. - Expired claims are replaced only with CAS against the observed object. - Release cannot delete a newer owner claim. - Stale live-state updates are rejected. - Normal branch history is unchanged by coordination operations.  A local bare-remote concurrency suite and a disposable create/read/delete probe against GitHub validate the underlying ref and blob behavior.

murdoch-development:
Closes #937

## Summary

- add blob-backed task coordination refs with explicit compare-and-swap updates
- add atomic claims, renewable leases, expired replacement, stale-write rejection, and ownership-safe release
- add `task claim`, `release`, `start`, `publish`, and `claims` CLI workflows
- preserve Markdown as the durable task record and keep coordination out of normal branch history
- document the workflow and remote/offline behavior

## Verification

- `bunx tsc --noEmit`
- modified TypeScript files pass Biome
- `bun run build`
- 9 focused coordination tests pass with 30 assertions
- disposable create/read/delete compatibility probe passed against GitHub

## Repository-wide checks

A full suite run progressed through the new tests and many existing suites, but the Windows checkout has unrelated baseline failures involving locked temporary files and the Claude-agent fixture. `bun run check .` likewise reports repository-wide CRLF formatting outside this change, while all modified TypeScript files pass Biome.

### DIFF
--- a/ADVANCED-CONFIG.md
+++ b/ADVANCED-CONFIG.md
@@ -46,6 +46,12 @@ Running `backlog config` with no arguments launches the interactive advanced wiz
 
 > **Note**: Set `remoteOperations: false` to work offline. This disables git fetch operations and loads tasks from local branches only, useful when working without network connectivity.
 
+### Remote task coordination
+
+When remote operations are enabled, Backlog can use blob-backed refs under `refs/backlog/tasks/*` for transient task coordination. These refs provide atomic claims, leases, and a live status/assignee overlay without adding commits to `main` or feature-branch history. Markdown task files remain the durable source of truth.
+
+Use `backlog task start BACK-123` to claim, assign, and move a task to the active status as one workflow. Use `backlog task claim`, `backlog task release`, `backlog task publish`, and `backlog task claims` for the individual operations. Claiming fails closed when the configured Git remote cannot be reached. Set `remoteOperations: false` to disable all remote coordination.
+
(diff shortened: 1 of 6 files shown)


murdoch-development:
Closing this draft because the implementation was opened before the required issue-first design review. The proposal is now tracked in #937 and the branch will remain available on the fork. We will wait for maintainer acceptance and scope guidance, then update the implementation, run the complete required checks.

PR #950, opened 2026-08-27
Count emoji as double-width in the TUI

### LINKED ISSUE #949: [Bug]: Emoji in task titles break TUI rendering (board column misalignment, garbled detail pane)
**Describe the bug**
Any emoji in a task title breaks TUI rendering, because the bundled `neo-neo-bblessed` measures every emoji as 1 terminal cell while modern terminals render them as 2:

1. **Kanban board** (`backlog board`): on rows whose title contains an emoji, everything after the emoji is shifted one cell right, so the column border on that row lands one column off — jagged/broken column edges.
2. **Task list** (`backlog task list`): when the details pane re-renders (e.g. moving the selection), blessed diffs against its incorrect model, and on any terminal row whose list entry contains an emoji the pane shows interleaved stale/new characters — visibly garbled text.

Root cause: `neo-neo-bblessed`'s `charWidth` delegates to the `eastasianwidth` npm package, whose data predates Unicode 9 (which reclassified emoji-presentation codepoints as Wide) — it returns Narrow for every emoji (`strWidth("🚀") === 1`). Independently, the wide-char layout regexes (`unicode.chars.wide`/`swide`) are hardcoded tables covering only CJK, so the layout engine never pads emoji either.

**To Reproduce**
1. `backlog task create "🚀 Emoji rocket task"` (plus a couple of plain-title tasks)
2. `backlog board` — the emoji row's column border is one cell right of the others
3. `backlog task list`, move the selection up/down — detail-pane rows that share a terminal row with an emoji list entry show mangled text

Confirmed via pty capture: with 3 columns at 100 cols, non-emoji rows place the border at column 33, emoji rows at 34.

**Expected behavior**
Emoji count as 2 cells so board borders stay aligned and detail-pane re-renders repaint cleanly.

**Environment**
- OS: Ubuntu (Linux 6.17), kitty + tmux; reproduces in any terminal that renders emoji 2 cells wide
- Backlog.md: current main (also 1.50.1)

**Additional context**
A fix is possible without waiting on the upstream blessed fork: patch `neo-neo-bblessed` via bun `patchedDependencies` so both `charWidth` and the layout regexes derive from `unicode-properties` East Asian Width data (already a dependency of the fork, with post-Unicode-9 data). I have this working, including VS16 sequences (`⚠️` = 2) and unchanged CJK/ASCII widths — PR to follow, happy to adjust the approach.

bjohas:
Fixes #949.

## Problem

`neo-neo-bblessed` measures every emoji as 1 terminal cell while modern terminals render 2:

- `charWidth` delegates to the `eastasianwidth` package, whose data predates Unicode 9 (which reclassified emoji-presentation codepoints as Wide) — `strWidth("🚀")` returns 1.
- Independently, the layout engine tags wide characters via hardcoded regexes (`unicode.chars.wide`/`swide`) that cover only CJK ranges, so emoji are never padded during layout either.

Result: on the kanban board, every row whose title contains an emoji shifts one cell right after the emoji, putting that row's column border out of line; in the task list, detail-pane re-renders interleave stale and new characters on any terminal row whose list entry contains an emoji.

## Fix

A dependency patch via bun `patchedDependencies` (`patches/neo-neo-bblessed@1.0.9.patch`), applied identically to `lib/unicode.ts` and both shipped `dist` bundles (the dist bundle is what actually runs):

- `charWidth` and the wide-char layout regexes now both derive from `unicode-properties` East Asian Width data — already a dependency of the fork, with post-Unicode-9 data — computed once at module load (~13ms), so there are no hand-maintained emoji tables and the two width mechanisms share one source of truth.
- VS16 sequences (`⚠️`, `❤️`) count 2 via a context-aware branch (a redundant VS16 after an already-wide emoji adds nothing).
- ASCII, CJK, and text-presentation widths (`⚠`, `🌡`) are unchanged.

Known limitation: `unicode-properties` ships pre-Unicode-14 data, so the newest emoji (e.g. 🫠 U+1FAE0) still measure 1. The right long-term home for this fix is the `neo-neo-bblessed` fork itself; this patch keeps Backlog.md correct in the meantime.

## Verification

- New test `src/test/tui-emoji-width.test.ts` (widths + layout-regex behavior).
- pty capture of `backlog board` at 100 cols with an emoji title: border cell lands at column 33 on emoji and non-emoji rows alike (was 34 vs 33 before).
- `bunx tsc --noEmit`, `biome check`, and the TUI-related test files pass.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -49,6 +49,9 @@
     "@biomejs/biome",
     "bun",
   ],
+  "patchedDependencies": {
+    "neo-neo-bblessed@1.0.9": "patches/neo-neo-bblessed@1.0.9.patch",
+  },
   "packages": {
     "@antfu/install-pkg": ["@antfu/install-pkg@1.1.0", "", { "dependencies": { "package-manager-detector": "^1.3.0", "tinyexec": "^1.0.1" } }, "sha512-MGQsmw10ZyI+EJo45CdSER4zEb+p31LpDAFp2Z3gkSd1yqVZGi0Ebx++YTEMonJy4oChEMLsxZ64j8FH6sSqtQ=="],
 
--- a/package.json
+++ b/package.json
@@ -97,5 +97,8 @@
     "@biomejs/biome",
     "bun",
     "node-pty"
-  ]
+  ],
+  "patchedDependencies": {
+    "neo-neo-bblessed@1.0.9": "patches/neo-neo-bblessed@1.0.9.patch"
+  }
 }
--- a/patches/neo-neo-bblessed@1.0.9.patch
+++ b/patches/neo-neo-bblessed@1.0.9.patch
@@ -1,0 +1,354 @@
+diff --git a/node_modules/neo-neo-bblessed/.bun-tag-fceb7ad038786fe4 b/.bun-tag-fceb7ad038786fe4
+new file mode 100644
+index 0000000000000000000000000000000000000000..e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
+diff --git a/dist/blessed.js b/dist/blessed.js
+index b33ef67fd6d2ba179d32fcbaf7207d966583aedb..0e7a3b967248153e8c8c642b3fb339e79d25776f 100644
+--- a/dist/blessed.js
++++ b/dist/blessed.js
+@@ -6905,6 +6905,50 @@ __export(unicode_exports, {
+   isSurrogate: () => isSurrogate,
+   strWidth: () => strWidth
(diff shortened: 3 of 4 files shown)


MrLesk:
Thanks @bjohas for the diagnosis and the regression test. The width fix moved upstream into neo-neo-bblessed 1.0.10, and #956 carries your commit with your authorship plus the dependency bump; your fork is organization-owned, so maintainer edits to this branch were not possible. Closing in favor of #956.

— Alex's agent

PR #956, opened 2026-08-29
BACK-646 - Count emoji as double-width in the TUI

### LINKED ISSUE #949: [Bug]: Emoji in task titles break TUI rendering (board column misalignment, garbled detail pane)

MrLesk:
Supersedes #950, which cannot receive maintainer edits because its fork is organization-owned. The original commit by @bjohas is preserved with their authorship, including their TUI regression test.

The width fix itself now lives in neo-neo-bblessed 1.0.10 (Unicode 16 Emoji_Presentation table driving both charWidth and the layout regexes, plus VS16 sequence handling), so this PR is the dependency bump to 1.0.10 in package.json, bun.lock, and bun.nix, removing the local patch from #950's first approach.

Closes #949.


MrLesk:
@codex review

PR #958, opened 2026-08-29
BACK-649 - Single-source task search config and filters in core

MrLesk:
Stage 1 of the task-search unification (BACK-649). `src/utils/task-search.ts` becomes the single owner of the searchable-text builder, the Fuse configuration, and the task filter predicates.

## What changed

**One search corpus.** `buildTaskSearchFields` produces exactly the fields `TASK_SEARCH_FUSE_OPTIONS` indexes, so the keys and the text behind them cannot drift. `SearchService` imports both and its duplicated `buildTaskBodyText` and inline Fuse config are deleted.

**Labels and assignees are searchable everywhere.** `SearchService`'s body text omitted them, so a query matching a label found the task via `task list --search` and MCP but not via `backlog search` or the web API. It no longer does.

**One filter implementation.** Four copies collapsed into `createTaskFilterMatcher`: `Core.applyTaskFilters`, `SearchService.applyTaskFilters` *and* its second copy `SearchService.matchesTaskFilters`, the CLI's `taskMatchesAllLabels`, and the MCP adapter's two hand-rolled label loops.

## labelMatch

Resolved explicitly: the shared predicate **defaults to `any`**, and callers passing an explicitly typed label list pass `all`.

Only one surface documents the semantics — the `task list --labels` help text, "require every comma-separated label" — and the MCP `labels` argument already required every label, so both pass `labelMatch: "all"`. Every remaining consumer is an interactive multi-select picker (TUI board and unified view, web label dropdown) where adding a label must *widen* the result set, so those keep `any`, as do the undocumented HTTP `label` parameters.

Two pre-existing bugs fell out of this:

- The plain/JSON `task list --labels` path filtered with a local all-labels helper *after* the query while `Core.applyTaskFilters` matched any label; one predicate now serves both.
- MCP label matching compared raw strings, so a task labelled `Backend` did not match `backend`. It is case-insensitive now, like every other surface.

MCP drafts also lose their hand-rolled filter chain. Draft assignee matching is now trimmed and case-insensitive (it was an exact `Array.includes`). The "search narrows drafts to the literal `Draft` status, listing them does not" quirk was preserved deliberately rather than silently changed.

## Out of scope

The local vs cross-branch corpus split (a deliberate v1.50 decision) is untouched, as are the TUI viewer fallback and `MilestonesPage` — those are BACK-650.

## Rebased onto BACK-643

BACK-643 (#924) added the `project` task attribute after this branch was cut, and it had added project filtering to **every** filter copy this PR deletes. A conflict resolution taking either side wholesale would have silently dropped project filtering from `backlog search --project`, `/api/search`, MCP, and the TUI.

The resolution folds it into the shared predicate instead: `TaskFilterOptions` gains `project`, `createTaskFilterMatcher` calls `matchesProjectFilter` (OR across the list; a task with no project never matches a non-empty filter), and every former call site keeps threading it — the `Core.queryTasks` `searchFilters` pass-through, the MCP task and draft paths including the draft status condition, the board, and the unified view.

## Wiring tests

An argument dropped at a call site is invisible to a predicate-level test, so the new `filter wiring across surfaces` block drives the real CLI and MCP surfaces. Each pin was mutation-verified:

| Mutation | Result |
| --- | --- |
| drop `labelMatch: "all"` from the MCP handler | 2 tests fail |
| drop `labelMatch: "all"` from CLI `baseFilters` | 1 test fails |
| drop the project check from the shared predicate | 2 tests fail |

All mutations were reverted after checking.

## Verification

Against a scratch project: `backlog search infrastructure`, `task list --search infrastructure`, and `GET /api/search?query=infrastructure` all return the same task, as do the same three surfaces for the assignee query `morgan`. `task list --labels backend,infrastructure` returns only the task carrying both; `--labels BACKEND` matches case-insensitively; `GET /api/tasks?label=backend&label=infrastructure` still returns both tasks, so the interactive picker keeps widening.

`bunx tsc --noEmit` clean, `bun run check .` clean, `bun test` 2563 pass / 0 fail — including fourteen cross-surface parity and wiring tests in `src/test/task-search-parity.test.ts`.

Net **-290 lines** of production code (+276 test).

### DIFF
--- a/backlog/tasks/back-649 - Single-source-task-search-config-and-filters-in-core.md
+++ b/backlog/tasks/back-649 - Single-source-task-search-config-and-filters-in-core.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-649
 title: Single-source task search config and filters in core
-status: To Do
-assignee: []
+status: In Progress
+assignee:
+  - '@claude'
 created_date: '2026-08-29 21:04'
+updated_date: '2026-08-29 22:15'
 labels:
   - cli
   - mcp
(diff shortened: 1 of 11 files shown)


MrLesk:
@codex review

MrLesk:
@codex review

PR #959, opened 2026-08-29
BACK-650 - Route TUI and milestone-page task search through the shared core search

MrLesk:
Stage 2 of the search unification, on top of BACK-649 (#958), which made `src/utils/task-search.ts` the single owner of the Fuse config, the searchable-text builder, and the task filter predicate. This routes the last two surfaces that still searched their own way.

## TUI task viewer

`src/ui/task-viewer-with-search.ts` kept two search paths: an in-memory `createTaskSearchIndex` when the caller passed tasks, and a `SearchService` branch otherwise — with the milestone, `labelMatch: "all"`, and readiness filters re-implemented by hand on top of the service's results because they could not be expressed through it.

It now builds one index over the loaded corpus however that corpus arrived, and `applyFilters` is a single `applyTaskFilters` call. The `hasActiveFilters` short-circuit went with it: an unset filter contributes no check, so the shared predicate already covers the no-filter case.

Two latent inconsistencies disappear as a side effect:

- the `SearchService` branch searched the whole ContentStore, while the unfiltered list rendered the prefix-filtered `allTasks` snapshot, so a query could surface a task the list itself would never show;
- it never re-indexed after an in-session edit, complete, or archive, because the rebuild was guarded on the index that branch did not use.

## Milestones page

`src/web/components/MilestonesPage.tsx` ran a third private Fuse over `id`/`title` with its own weights (0.55/0.45). Deleted; the bucket corpus goes through the shared index.

Shared index rather than `/api/search`: the page filters buckets already built from its `tasks` prop, so an API round trip would only be reconciled back to the same corpus by id, and the filter stays synchronous — no debounce, no loading state.

## User-visible behavior changes

Both are on the milestones page, and both are the shared semantics arriving:

1. **Exact-id-wins is gone.** The page used to short-circuit to exact id matches; a full-id query now also returns the fuzzy id neighbours (searching `task-404` in the test's four-task corpus also matches `task-101`, `task-202`, `task-303`). This is what the TUI and the sidebar already do with the shared threshold — the sidebar just hides it by ranking and slicing to 5, whereas this page uses the result as a set. If the noise is not acceptable, the fix belongs in the shared config for every surface rather than as a local exception here; happy to open that as a follow-up.
2. **Search widened from id + title to the full shared corpus** — description, acceptance criteria, plan, notes, comments, labels, assignees. The placeholder changed from "Search by task ID or title" to "Search tasks" to stop understating it.

Nothing else changes: the TUI's visible filter behavior is identical, and `Board.tsx`'s three-line client-side label display filter was left alone as previously decided — routing the milestones page did not make folding it free.

## Tests

- `src/test/web-milestones-page-search.test.tsx`: the rendered task set equals `createTaskSearchIndex(corpus).search({ query })` for the same query, across six queries including an id query and a no-match query.
- `src/test/task-search-parity.test.ts`: milestone-title, no-milestone, and readiness filters return the same set with and without a query — the three filters the TUI used to hand-roll, where a query was exactly what made them drift.

Combined with the existing shared-index/`SearchService` parity tests in that file, this pins both surfaces to the same results as the CLI, MCP, and web API.

## Verification

- `bunx tsc --noEmit` clean
- `bun run check .` clean
- `bun run test` — 2567 pass, 0 fail

One environment note: the worktree needed a `bun install` first. The shared `node_modules` had `neo-neo-bblessed` 1.0.9 while main locks 1.0.10, which failed three unrelated `tui-emoji-width` tests. No dependency change in this PR.

### DIFF
--- a/backlog/tasks/back-650 - Route-TUI-and-milestone-page-task-search-through-the-shared-core-search.md
+++ b/backlog/tasks/back-650 - Route-TUI-and-milestone-page-task-search-through-the-shared-core-search.md
@@ -1,10 +1,11 @@
 ---
 id: BACK-650
 title: Route TUI and milestone-page task search through the shared core search
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@claude'
 created_date: '2026-08-29 21:04'
-updated_date: '2026-08-29 21:04'
+updated_date: '2026-08-29 23:00'
 labels:
   - tui
   - web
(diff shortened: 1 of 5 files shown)


MrLesk:
@codex review

PR #967, opened 2026-08-30
BACK-657 - Make the TUI acceptance-criteria bar degrade gracefully without Block Element glyphs

MrLesk:
## Summary
- The list/board acceptance-criteria progress bar was built from raw U+2588 FULL BLOCK and U+2591 LIGHT SHADE. blessed only guarantees fallback for DEC Special Graphics (box drawing) via the ACS charset, so a terminal font lacking Block Elements drew the bar as blank cells (maintainer-observed: a 9/10 bar rendered empty), and without a UTF-8 locale every cell became `?`.
- The bar now uses plain ASCII: `#` for filled cells and `-` for empty ones, e.g. `[######----] 4/7`.

## Why ASCII instead of ACS-routed glyphs
The only fill-like glyph in the DEC Special Graphics set is the checkerboard U+2592 (ACS `a`). On terminals with broken or absent ACS, blessed falls back to its `utoa` table, which maps U+2592 to a **space** — the bar would vanish on exactly the degraded terminals this fix targets, and it reads as a half-shade rather than filled. ASCII stays below `~`, so it bypasses every charset translation in `neo-neo-bblessed`'s screen draw and renders identically on all terminals, fonts, and locales.

## Scope
- Only `src/ui/acceptance-criteria-progress.ts` and its test change (the web `AcceptanceCriteriaProgress` component is tracked separately as BACK-652; `src/ui/board.ts` untouched to avoid overlap with #945).

## Verification
- `bunx tsc --noEmit` clean
- `bun run check .` clean
- `bun run test`: 2568 pass, 0 fail
- Tests pin the exact emitted characters and an ASCII-only shape (`/^\[[#-]*\] \d+\/\d+$/`) across widths and ratios.

### DIFF
--- a/backlog/tasks/back-657 - Make-the-TUI-acceptance-criteria-bar-degrade-gracefully-without-Block-Element-glyphs.md
+++ b/backlog/tasks/back-657 - Make-the-TUI-acceptance-criteria-bar-degrade-gracefully-without-Block-Element-glyphs.md
@@ -3,9 +3,11 @@ id: BACK-657
 title: >-
   Make the TUI acceptance-criteria bar degrade gracefully without Block Element
   glyphs
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@claude'
 created_date: '2026-08-30 13:19'
+updated_date: '2026-08-30 15:40'
 labels:
   - tui
   - bug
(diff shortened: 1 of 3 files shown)


MrLesk:
@codex review

PR #969, opened 2026-08-30
BACK-658 - Resolve dependency targets in completed and archived tasks

### LINKED ISSUE #942: [Bug]: task edit --depends-on rejects ids in completed/, making a dependency list permanently uneditable
**Describe the bug**

`backlog task edit <id> --depends-on <ids>` — and the MCP `task_edit(dependencies=[...])` parameter above it — resolves every supplied id against `backlog/tasks/` only. A task that has been completed lives in `backlog/completed/`, so it is reported as nonexistent and the **entire write is refused**:

Two properties combine into something worse than either alone:

1. **The check is location-based, not status-based.** A task with `status: Done` that still sits in `backlog/tasks/` is accepted; the identical task is rejected the moment it moves to `completed/` (or `archive/`).
2. **`--depends-on` replaces rather than appends** — there is no `--depends-on-add` / `dependenciesAdd` — so a caller must supply the union of the existing list plus the new entry.

Together: once *any* single entry in a task's dependency list has been completed, **every union the caller can construct contains a rejected id**, and the dependency list can never be edited again by any supported means.

**To Reproduce**

1. Create `TASK-100` and `TASK-200`, and a task `TASK-A` with `dependencies: [TASK-100]`.
2. Complete `TASK-100` so its file moves to `backlog/completed/`.
3. Try to add `TASK-200` while preserving the existing entry:
4. See: `The following dependencies do not exist: TASK-100.` — the whole write is refused and the list is unchanged.

Control: the same call with only ids that still live in `backlog/tasks/` succeeds, which is what isolates this to the *location* of the referenced file rather than its status.

**Expected behavior**

A dependency on completed work is the normal end state of a predecessor, so `TASK-100` should resolve and the write should land as `[TASK-100, TASK-200]`.

Concretely, either fix would restore recoverability, and both together would be ideal:

1. Resolve dependency ids across `backlog/tasks/` **+ `backlog/completed/` + `backlog/archive/`**.
2. Add an append parameter (`--depends-on-add` on the CLI, `dependenciesAdd` on the MCP tool) so a caller never has to restate a list it cannot round-trip.

**Environment**
- OS: Windows 11 x64
- backlog.md: 1.50.1 (also reproduced on 1.44.0)
- Installed globally via bun (`backlog.md-windows-x64`), driven both through the CLI and the MCP server

**Additional context**

The only way to force the write through is to **drop the completed entry** — deleting the record of a predecessor that really did run, in order to record one that has not. That is provenance destruction, so in practice the list is simply frozen.

In our repo this produced a concrete correctness problem rather than an inconvenience: a task's declared dependencies were all `Done`, so tooling that resolves predecessors from frontmatter reported it **ready to action**, while its real, still-open predecessor could only be recorded in prose — because the edit that would have declared it is the edit that is refused.

Separately, and as a thank-you: the sibling `task list --milestone` behaviour (fuzzy match on milestone *title*, so an `m-NNN` id silently returned a different milestone's task set) **appears fixed in 1.50.1**. We re-measured 53/53 populated milestones agreeing with frontmatter, against 49/49 mis-resolving previously. The dependency resolver above is unchanged between 1.44.0 and 1.50.1.

MrLesk:
Closes #942. Implements BACK-658.

## Problem

`validateDependencies` built its known-ID corpus from working-copy tasks plus drafts only, so a dependency on a task living in `backlog/completed/` or the archive was refused at write time — even though Done is the normal end state of a predecessor. Combined with the replace-only semantics of `--depends-on`, a task whose predecessor completed could no longer have its dependency list edited at all.

## Change

- `validateDependencies` (src/utils/task-builders.ts) now resolves dependency targets across working-copy tasks, drafts, completed, and archived records, using the same shared identity rules (`taskIdsEqual` / `canonicalTaskId`) as before.
- The `core.loadTaskById` working-copy ambiguity check is unchanged: its identity index already covers completed tasks, and archived-only IDs resolve to null there exactly like drafts — the call exists only to catch several files claiming one ID.
- Validation only. Readiness and dependency-graph semantics for completed/archived targets are untouched, and nothing here overlaps the planned self/cycle checks (BACK-656).
- Unknown IDs are still rejected; ambiguous identities still fail closed, including the bare-numeric case spanning the task and draft counters and duplicate identities across active/completed/archived stores.

CLI, MCP, and web all funnel through `createTaskFromInput`/`updateTaskFromInput`, so the one corpus fix covers all three surfaces.

## Tests

New coverage in `src/test/dependency.test.ts`:

- completed target accepted at create and edit time
- archived target accepted at create and edit time
- re-editing the dependency list of a task whose predecessor completed (the #942 scenario, replace semantics revalidating the completed entry)
- draft target still accepted, unknown ID still rejected
- duplicate identity across active and completed stores fails closed

The existing bare task/draft numeric ambiguity test (`cli-custom-prefix-id-resolution.test.ts`) stays green.

## Verification

- `bunx tsc --noEmit` clean
- `bun run check .` clean
- `bun run test` green

### DIFF
--- a/backlog/tasks/back-658 - Resolve-dependency-targets-in-completed-and-archived-tasks.md
+++ b/backlog/tasks/back-658 - Resolve-dependency-targets-in-completed-and-archived-tasks.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-658
 title: Resolve dependency targets in completed and archived tasks
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@Fable'
 created_date: '2026-08-30 15:23'
+updated_date: '2026-08-30 15:43'
 labels:
   - cli
   - mcp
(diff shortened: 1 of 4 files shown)


MrLesk:
@codex review

PR #970, opened 2026-08-30
BACK-659 - Include grandchild subtasks in board export grouping

### LINKED ISSUE #944: [Bug]: board export omits grandchildren and deeper subtasks
**Describe the bug**

`backlog board export` silently omits grandchildren and deeper same-status subtasks.

In a three-level hierarchy, all tasks are present in `task list --json`, and `task view` reports the grandchild's parent correctly. The export exits successfully with empty stderr, but the generated Markdown contains only the parent and direct child.

**To Reproduce**

From a fresh directory:

```sh
backlog init "Nested Board Export Repro" \
  --defaults \
  --integration-mode none \
  --no-git

backlog task create "Parent" --type task --plain
backlog task create -p TASK-1 "Child" --type task --plain
backlog task create -p TASK-1.1 "Grandchild" --type task --plain

backlog task list --json
backlog task view TASK-1.1.1 --json
backlog board export board.md --force
```

The read commands show all three IDs and `TASK-1.1.1` has `parentTaskId: "TASK-1.1"`:

```text
TASK-1
TASK-1.1
TASK-1.1.1
```

The generated board contains only:

```markdown
| **TASK-1** - Parent |  |  |
| └─ **TASK-1.1** - Child |  |  |
```

`TASK-1.1.1` is absent.

**Expected behavior**

The exported board should include every task in the hierarchy, including grandchildren and deeper subtasks, without silently dropping valid tasks. The exact visual indentation can follow the existing board conventions.

**Environment**

- OS: macOS 26.6.2, arm64
- Node version: v24.19.0
- Backlog.md version: 1.50.1

**Additional context**

- Reproduced twice in isolated filesystem-only projects.
- All involved commands exited successfully; read and export stderr were empty.
- The behavior is consistent with the current one-level grouping path in [`src/board.ts`](https://github.com/MrLesk/Backlog.md/blob/988d27fe4d0c556042088f71035d17002c41ac54/src/board.ts#L90-L140), which appends direct children of top-level tasks but does not emit descendants of those children.
- The original subtask board work in #54 states that subtasks work across board export operations; no documented hierarchy-depth limit was found.
- Open PR #896 touches direct-child sorting in the same file but does not address omission of deeper descendants.

MrLesk:
Fixes #944.

## Problem
The markdown board export (`backlog board export`) built each status column as top-level tasks plus their direct children only. A subtask whose parent is itself a subtask was collected into the children map under the intermediate subtask's id — which was never walked — so grandchildren silently disappeared from the exported board.

## Fix
Flatten each column depth-first over the existing children map (`pushWithChildren` recursion in `src/board.ts`) instead of one level deep. The per-level ID-ascending sort is unchanged, and grandchildren render with the existing `└─ ` subtask shape.

## Proof
- New test pins the nested chain (parent → child → grandchild) with exact expected output; it fails on pre-fix code and passes now.
- New exact-output flat parent/child test passes against both pre-fix and post-fix code (verified via `git stash`), proving flat export output is byte-identical.
- `bunx tsc --noEmit`, `bun run check .`, and `bun test src/test/board.test.ts` (10/10) pass. Full-suite run shows only the 3 pre-existing `tui-emoji-width` failures that also fail on clean main in this environment.

Backlog task: BACK-659 (included in this PR with checked ACs and final summary).

### DIFF
--- a/backlog/tasks/back-659 - Include-grandchild-subtasks-in-board-export-grouping.md
+++ b/backlog/tasks/back-659 - Include-grandchild-subtasks-in-board-export-grouping.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-659
 title: Include grandchild subtasks in board export grouping
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@claude'
 created_date: '2026-08-30 15:23'
+updated_date: '2026-08-30 15:45'
 labels:
   - cli
   - bug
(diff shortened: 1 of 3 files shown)


MrLesk:
@codex review

PR #972, opened 2026-08-30
BACK-660 - Reject nested section markers in notes and fix append truncation

### LINKED ISSUE #932: task edit --notes nests SECTION:NOTES markers when the payload contains them, silently hiding the note from task view
## Summary

`backlog task edit <id> --notes <text>` wraps the payload in a fresh `SECTION:NOTES:BEGIN/END` pair without reconciling markers already present in that payload. If the text you pass happens to contain the section markers — which it does whenever you implement "append to notes" as read-modify-write — the file ends up with **nested markers**, and everything after the first `END` becomes invisible to `task view`.

The write reports success: exit `0`, `Updated task TASK-1`. The file grows. `grep` finds the new text. Only the renderer disagrees, so a naive "did my write land?" check (bytes grew / text present) passes while the note is effectively lost.

Version: **1.50.1** (also reproduced on 1.48.0). macOS, arm64, installed via `npm i -g backlog.md`.

## Reproduction

```bash
mkdir repro && cd repro && git init -q .
backlog init "ReproTest" --defaults --no-git --integration-mode none --backlog-dir .backlog
export BACKLOG_CWD="$PWD/.backlog"

backlog task create "repro" -d "seed description"
backlog task edit 1 --notes "FIRST NOTE"

F=$(find .backlog/tasks -name 'task-1*.md')
# Read the notes body back the way a script would — this includes the marker lines.
BODY=$(sed -n '/^## Implementation Notes/,$p' "$F" | tail -n +2)
backlog task edit 1 --notes "$BODY

SECOND NOTE"

grep -c 'SECOND NOTE' "$F"              # 1  — it IS in the file
backlog task view 1 | grep -c 'SECOND NOTE'   # 0  — but not in the render
```

Measured: file `362 B` → `431 B`, marker counts `BEGIN=1 END=1` → `BEGIN=2 END=2`, `exit=0` throughout.

Resulting layout:

```
19:## Implementation Notes
21:<!-- SECTION:NOTES:BEGIN -->
22:<!-- SECTION:NOTES:BEGIN -->
24:<!-- SECTION:NOTES:END -->     <- renderer stops here
27:<!-- SECTION:NOTES:END -->
```

## Expected

Either reconcile the markers (replace the marked region rather than emit a new pair beside it), or reject a `--notes` payload containing `SECTION:*` markers with a clear error. Silently producing an unreadable file is the worst of the three.

## Second symptom: a clean write does not repair it

Once a file is nested, writing clean notes over it (payload containing **no** markers) does not restore a valid structure — it leaves an **orphan `END` with the old text stranded above the `## Implementation Notes` header**, outside every section:

```
19:SECOND NOTE                  <- stranded, above the header, in no section
20:<!-- SECTION:NOTES:END -->   <- orphan
22:## Implementation Notes
24:<!-- SECTION:NOTES:BEGIN -->
25:REPAIRED BODY
26:<!-- SECTION:NOTES:END -->
```

`BEGIN=1 END=2`. The new body renders, but the displaced text stays in the file indefinitely. Recovering it needs hand editing. Same on 1.48.0 and 1.50.1.

## Why this bites in practice

`--append-notes` exists, but it isn't a safe substitute for every case: it reads the existing notes up to the first end-marker **substring** after `BEGIN`, so a notes body that *quotes* the terminator gets truncated by the additive flag (measured separately at byte level: 392 B → 375 B, exit 0). That pushes callers toward read-modify-write with `--notes`, which is exactly the path that nests.

Related but distinct: #930 (`--notes` collapsing blank lines inside fenced code blocks).

## Note on marker checks

If you add a validation for this, matching `SECTION:NOTES:BEGIN` as a substring will produce false positives on any task whose own prose mentions the markers. Worth anchoring on the whole comment line.

Separately, a structural rule alone can over-report: in a store of 2,996 tasks I found 4 files with malformed marker sequences that nonetheless **render in full**, because the renderer appears to fall back to dumping a section raw when the region doesn't open with `BEGIN`. So "more than one BEGIN" is not by itself equivalent to "content is unreachable".

MrLesk:
Fixes #932. Backlog task: BACK-660.

## Problem

`task edit --notes` wrapped payloads in a fresh `SECTION:NOTES:BEGIN/END` pair without reconciling marker lines already inside the payload, so the read-modify-write pattern nested the markers and hid everything after the inner `END` from `task view` — while the write reported success. Separately, section extraction stopped at the first end-marker *substring*, so notes that merely mention the terminator inline were truncated on read, and `--append-notes` persisted the loss. A clean rewrite over a nested file stranded the displaced text above the section header instead of repairing it.

## Fix

**Reject, not strip.** New section input (description, plan, notes, final summary — including the append flags) that contains the target section's *own* sentinel marker as a whole line now fails with a clear error naming the marker, before any write. Rationale: this content is the durable record, and silently altering what the caller asked to store is worse than a retryable error; comment bodies already reject reserved markers the same way. The check is per-family and whole-line only, so inline mentions, indented marker lines, and other sections' markers (including the fenced cross-family examples locked in by BACK-637) remain storable. Validation lives at the shared core input boundary (`createTaskFromInput` + `applyTaskUpdateInput`), so CLI, MCP, web server, and TUI create all agree; TUI raw-file editing remains the human escape hatch.

**Real boundaries.** The five sentinel-path regexes in `structured-sections.ts` are replaced by one line-anchored, depth-aware block scanner (`findSentinelBlocks`) shared by extraction, section ranges, insert-position lookup, and strip. Markers count only as whole lines (per the issue's own guidance), so inline substrings can never terminate a section early and append no longer truncates. Same-family `BEGIN` lines are depth-counted, so a pre-existing nested-marker file resolves to its full interior: `task view` shows every note line, and a clean `--notes` rewrite strips the whole nested region — repairing the file instead of stranding text. `tokenizeKnownSentinels` is line-anchored the same way, closing the equivalent inline-mention hazard for AC/DOD/comment masking.

## Verification

- New suite `src/test/section-marker-safety.test.ts` (9 tests): rejection across create/edit/append with readable files after failure, inline/indented/fenced-foreign markers still round-tripping, append with marker-like substrings (byte-level no-shrink), nested-file readability in `task view`, repair on clean rewrite, and round-trip stability. 8 of 9 fail on pre-fix code.
- BACK-637 fence round-trip suite untouched and green.
- `bunx tsc --noEmit`, `bun run check .`, and full `bun test` green — except 3 `tui-emoji-width` failures that reproduce identically on a clean `origin/main` checkout (environment-specific, unrelated).

### DIFF
--- a/backlog/tasks/back-660 - Reject-nested-section-markers-in-notes-and-fix-append-truncation.md
+++ b/backlog/tasks/back-660 - Reject-nested-section-markers-in-notes-and-fix-append-truncation.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-660
 title: Reject nested section markers in notes and fix append truncation
-status: To Do
-assignee: []
+status: Done
+assignee:
+  - '@Claude'
 created_date: '2026-08-30 15:23'
+updated_date: '2026-08-30 15:48'
 labels:
   - cli
   - bug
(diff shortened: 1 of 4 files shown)


MrLesk:
@codex review

PR #974, opened 2026-08-30
BACK-664 - Read the Backlog overview once per conversation instead of per request

MrLesk:
The shipped agent nudge told agents to run `backlog instructions overview` for every user request, so agents re-read static content many times per conversation (maintainer-observed). The trigger is now once at the beginning of each conversation, re-reading only if not yet read in the current conversation. Lifecycle guide triggers are unchanged. Single-source change in cli-agent-nudge.md plus this repo's own AGENTS.md instance and the init test.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -108,7 +108,7 @@ are found, the commit will be blocked until fixed.
 
 This project uses Backlog.md for task and project management.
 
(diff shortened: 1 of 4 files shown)


MrLesk:
@codex review

PR #983, opened 2026-08-31
BACK-641 - Add `backlog task dependencies` command with TUI and plain graph views

MrLesk:
## Summary

Adds `backlog task dependencies <id>`, a dedicated view of a task's dependency graph over the shared model from BACK-548. The command consumes the existing `TaskDetail` derived graph (`loadTaskCorpus` / `withDependencyGraph`) — no new computation path, per-surface loader, or endpoint.

- **Interactive TUI (default in a TTY)**: renders the same tree the plain serializer prints; ↑↓/j/k move between task rows (headings and separators are skipped; wheel or heading clicks snap to the nearest task row), Enter **or a single mouse click** re-roots the view on the selected task, Esc/q exits. Unresolved identities (`unknown task ID` / `ambiguous task ID`) and cycle markers render exactly as everywhere else and are never navigated into. Titles are `{open}`/`{close}`-escaped via the shared TUI label helper, which `task-viewer-with-search` now imports instead of keeping a private copy.
- **`--plain`**: the identical `Dependency Graph:` block that `task view --plain` renders, through the one shared serializer (`formatDependencyGraphLines`); an isolated task prints a one-line `... has no dependencies and no dependents.` message instead of an empty graph.
- **`--json`**: `{ schemaVersion: 1, kind: "task-dependencies", dependencyGraph: { root, nodes, edges } }` — the normalized graph representation defined by BACK-548, shared with task-view JSON via one `toDependencyGraphJson` helper.
- **MCP**: read-only `task_dependencies` tool (`{ id }`) registered next to `task_view`; its text output is byte-identical to the CLI `--plain` output (asserted in tests).
- **Docs**: CLI help schema plus the shipped CLI/MCP instruction guides (`task-execution`, MCP overview/tools) document the command, flags, and edge-direction semantics.

Fix folded in: neo-neo-bblessed 1.0.10 never activates its lazy mouse parser for the screen's own listener registration, so clicks were dead; the view calls the idempotent `program.bindMouse()` explicitly. The same latent issue likely affects other `mouse: true` TUI views — flagged separately as follow-up rather than expanded here.

## Test evidence

- `bunx tsc --noEmit` — clean
- `bun run check .` — clean
- `bun run test` — **2798 pass / 0 fail** (8 skip)
- New `src/test/cli-task-dependencies.test.ts`:
  - CLI `--plain`: chains in both directions, diamonds (`(shown above)`), cycles (`(cycle)`), missing + ambiguous references rendered explicitly and fail closed, isolated one-line message, exact layout equality with the `task view --plain` graph block
  - CLI `--json`: full normalized nodes/edges shape, unresolved states, `--json`+`--plain` rejection, missing task exits 1
  - Serializer entries: node reference on every task line, none on headings/separators; TUI labels color unresolved/completed nodes and escape blessed tags without changing the shared wording
  - MCP: `task_dependencies` output equals CLI `--plain` byte-for-byte; isolated and missing-task cases
- `src/test/mcp-server.test.ts` tool listings include the new tool
- Manual TUI verification through a sized `expect` PTY harness: initial render, Enter re-root (BACK-641 → BACK-548), j/k movement, X10 mouse click re-root (3/3 runs), q exit and screen restore


MrLesk:
@codex review

PR #985, opened 2026-09-01
BACK-674 - Sort the TUI list view through the shared task ID comparator

### LINKED ISSUE #953: [Bug]: Sort order in task view of the backlog board is wrong
**Describe the bug**
In the kanban view the sub-tasks seems to be sorted according to the numbers -p assigns them, which is an exected order, however in the list view they seem to be sorted alphabetically, which put ABC-1.10 in front of ABC-1.2.

**To Reproduce**
1. Create more than 9 subtasks to a task with -p
2. Observe in the kanban view that 1.10 goes after 1.9
3. Observe in list view that 1.10 goes in front of 1.2

**Expected behavior**
Sort order should be either configurable and persisted (preferably) or consistent. It makes sense to sort in numerical order like the kanban view does rather then alphabetically like list view does.

**Environment**
- OS: Windows 11
- Node version: 24, but this is a bun compiled exe, so does it matter?

**Additional context**

Image

MrLesk:
Fixes #953

## The report

With more than nine subtasks, the board orders them numerically (1.9 then 1.10) while the task list orders them alphabetically (1.10 before 1.2).

## What was actually wrong

Reproduced on a scratch project (custom `ABC` prefix, one parent, eleven subtasks created with `-p`) and compared every surface before changing anything:

| surface | order |
| --- | --- |
| `task list --plain` / `--json` | correct |
| web task list | correct |
| TUI board columns | correct |
| `filesystem.listTasks()`, `Core.queryTasks()` | correct |
| `Core.loadTasks()` | **`ABC-1.1, ABC-1.10, ABC-1.11, ABC-1.2, …`** |

`Core.loadTasks()` is the corpus `backlog board` hands to the unified view. The board re-sorts it into columns, but the TUI task list renders the array as given, so pressing Tab from the board showed the alphabetical order in the reporter's screenshot. That matches the issue title, "Sort order in task view of the backlog board".

The TUI list view itself is not the defect. Every other entry point into it (`task list`, `draft list`, `search`, `task view`) already hands it a deliberately sorted array, and `task list --sort` and its priority default live there — sorting inside the view would have overridden those intentional orderings.

Root cause: `TaskIdentityIndex.getTasks()` ordered identity groups with `left.id.localeCompare(right.id)`, a second ad-hoc ID comparator. `ContentStore` masked it by re-sorting with `sortByTaskId`; `Core.loadTasks()` returns `getTasks()` order verbatim.

## The change

`getTasks()` now sorts groups with the shared `compareTaskIds`, so the duplicate comparator is removed rather than another sort added. One line plus the import.

No deliberate ordering was touched: the board still orders by ordinal with parent/subtask grouping, `task list` keeps `--sort` and its priority default, and the web list keeps its descending default. `backlog board export` reads the same corpus, so its subtask rows come out numeric now too.

## Tests

`src/test/subtask-ordering-consistency.test.ts` builds `TASK-1` with subtasks `1.1`–`1.11` (created in reverse) plus `TASK-2` and `TASK-11`, then asserts the board corpus, every corpus reader, the TUI board columns, the comparator the web list sorts by, and CLI plain and JSON all produce one identical ID order. A `TaskIdentityIndex` unit test guards the changed line directly. Both fail against the old comparator and pass after. `src/ui/board.ts` exports `prepareBoardColumns` so the board assertion reads real board columns instead of re-deriving them.

## Out of scope

The report also asked for a configurable and persisted sort order. Not built: `backlog task list --sort` already exists, and the defect here is consistency.

## Visible output change

`backlog board export` reads the same corpus, so exported markdown boards now list subtasks numerically (1.9 before 1.10) instead of alphabetically. Maintainer-confirmed as correct.

### DIFF
--- a/backlog/tasks/back-674 - Sort-the-TUI-list-view-through-the-shared-task-ID-comparator.md
+++ b/backlog/tasks/back-674 - Sort-the-TUI-list-view-through-the-shared-task-ID-comparator.md
@@ -5,7 +5,7 @@ status: In Progress
 assignee:
   - '@claude'
 created_date: '2026-09-01 17:16'
-updated_date: '2026-09-01 17:40'
+updated_date: '2026-09-01 17:44'
 labels: []
 dependencies: []
 ordinal: 306000
@@ -71,6 +71,8 @@ Validation: bunx tsc --noEmit clean, bun run check . clean, bun run test green a
 Also observed: backlog board export reads the same corpus, so exported markdown boards now list subtasks numerically too. Verified on the reproduction project.
 
 PR: https://github.com/MrLesk/Backlog.md/pull/985
+
+Accepted behavior (maintainer decision, 2026-09-01): a task created while the TUI list view is open still appends to the bottom rather than landing in sort position. The view is deliberately ignorant of ordering — callers hand it an already-sorted corpus, so honoring the active order on insert would mean threading sort state into the view and risks clobbering --sort and the priority default. The placement corrects itself when the view is reopened. Do not 'fix' this without an explicit decision.
 <!-- SECTION:NOTES:END -->
 
 ## Final Summary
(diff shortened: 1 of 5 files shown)


MrLesk:
@codex review

PR #993, opened 2026-09-02
BACK-670 - Remove the standalone task dependencies command and its TUI

MrLesk:
Closes BACK-670.

The dependency graph is a derived field on the task detail and is already rendered by every surface — CLI `--plain`, `--json`, the TUI viewer, the web modal, and MCP `task_view`. The standalone `backlog task dependencies` command shipped later as a second window onto the same data, minus the task itself, and its interactive view contradicted its own footer (`[Enter/Click] Open task` while Enter re-rooted the graph). It is removed rather than grown. Corpus-level dependency exploration remains tracked separately as BACK-671.

## Removed

- `backlog task dependencies` in `src/cli.ts`: the command, its `addHelpSchema` entry, and its slot in the `taskCmd --json` allowlist.
- `src/ui/dependencies-tui.ts`, the interactive graph view.
- The MCP `task_dependencies` tool: definition and registration in `src/mcp/tools/tasks/index.ts`, the handler in `handlers.ts`, `taskDependenciesSchema` in `schemas.ts`, and the schema re-export.
- The formatters that had no other caller left: `taskDependenciesJson` (`src/formatters/json-output.ts`) and `formatTaskDependenciesPlainText` (`src/formatters/task-plain-text.ts`).
- The four references in shipped agent instructions: `cli-instructions/task-execution.md`, `mcp/overview.md`, `mcp/overview-tools.md`, `mcp/task-execution.md`.
- `src/test/cli-task-dependencies.test.ts`, and `task_dependencies` from the two tool-list expectations in `src/test/mcp-server.test.ts`.

57 insertions against 643 deletions across `src/`.

## Kept

`formatDependencyGraphEntries` already lived in `src/formatters/dependency-graph-text.ts` and did not move. `formatDependencyNodeTuiLabel` and its private `escapeBlessedTags` helper moved there verbatim, next to the plain `formatDependencyNodeLabel` whose wording they wrap, so the plain and colored labels cannot drift apart. Its only consumer, `src/ui/task-viewer-with-search.ts`, already imported `formatDependencyGraphLines` from that module, so the move removes an import rather than adding one.

The two tests from the deleted file that covered surviving code — entries pair each task line with its node and leave headings unattached; the TUI label colors unresolved and completed nodes and escapes braces — moved into the existing `dependency graph text` block in `src/test/dependency-graph.test.ts`.

## Verification

Task detail output is unchanged on every surface.

A scratch project covering a chain, a diamond with a `(shown above)` repeat, a cycle, an unknown ID, an ambiguous ID, a completed dependency and an isolated task was captured before and after the change against the same project directory: `task view <id> --plain`, `task view <id> --json` and `task <id> --plain` for all ten resolvable tasks, plus `task list --plain`, `task edit 4 --plain` and `doctor`. `diff` reports no difference — 1635 lines, 41749 bytes, md5 `96872cbb0374cd5e19a7112e320b68d7` on both sides. The snapshot was first proven stable across two consecutive pre-change runs, so the diff measures the code change rather than run-to-run noise.

The TUI and web graph sections stay covered by `src/test/dependency-graph-surfaces.test.tsx` (which renders the real viewer body, including the blessed-tag escaping and the colored unresolved/completed labels) and `src/test/web-dependency-graph-section.test.tsx`.

`bunx tsc --noEmit` and `bun run check .` pass. `bun run test`: 2846 pass, 8 skip, 0 fail, plus the known local-only `src/test/content-store.test.ts` flake, which also appears on unmodified `origin/main` and passes when that file is run on its own (67 pass on both).


MrLesk:
@codex review

PR #995, opened 2026-09-02
BACK-684 - Move the acceptance-criteria ring into the web card header

MrLesk:
## Summary

On web board cards the acceptance-criteria ring rendered on its own line under the title, so a 12px ring and a `3/5` label cost a full row of card height on every In Progress card with criteria. The ring is metadata, and the card already has a metadata row: the header with the task id, type and project badges on the left and the priority badge on the right.

This moves the ring into that header row and removes the standalone line, so cards with progress get **shorter instead of taller**.

## What changed

`src/web/components/TaskCard.tsx`:

- `getPriorityBadge` is now called once into a `priorityBadge` value.
- The header's right side is a `flex shrink-0 items-center gap-2` group that renders **only** when there is progress or a priority badge — a card with neither still renders nothing beside the id.
- Inside the group the ring comes first and the priority chip last, so the chip stays flush right exactly where cards without progress already put it. The id/type/project group keeps `min-w-0` and remains the only side that gives way.
- The standalone `<AcceptanceCriteriaProgress ... className="mt-2" />` line under the title is gone.

`AcceptanceCriteriaProgress` itself and the task list usage (`density="list"` in `TaskList.tsx`) are unchanged.

## Verification

Rendered on a local browser board in both light and dark themes:

- `BACK-636` (Medium priority, 0/4) renders `BACK-636 … ◯ 0/4 Med` on one header line; `BACK-684` (no priority, 0/4) renders the ring flush right. No progress node outside the header on either card, and both titles sit directly under the header.
- Measured in the DOM: re-inserting the old standalone line under the title grew the `BACK-636` card from **101px to 129px**, so the header placement saves 28px (~22%) per card with progress.
- Squeezed a card carrying a type badge from 260px down to 130px: the header stayed a single 20px line with zero overflow — the right group holds its width and the left group truncates instead of wrapping.

Tests — `src/test/web-task-acceptance-progress.test.tsx` gains four card cases: ring plus priority badge (ring inside the header row, exactly one progress node, badge still last), ring without priority, priority without progress, and neither. The two placement cases fail against the pre-change component.

Gates:

- `bunx tsc --noEmit` — clean
- `bun run check .` — clean (425 files)
- `bun run test` — 2859 pass, 8 skip, 1 fail: `ContentStore > refreshes completed identity state when the completed corpus changes`, a known local-only flake that passes when its file is run alone (67/67) and is unrelated to this change.

Task: BACK-684

### DIFF
--- a/backlog/tasks/back-684 - Move-the-acceptance-criteria-ring-into-the-web-card-header.md
+++ b/backlog/tasks/back-684 - Move-the-acceptance-criteria-ring-into-the-web-card-header.md
@@ -1,9 +1,11 @@
 ---
 id: BACK-684
 title: Move the acceptance-criteria ring into the web card header
-status: To Do
-assignee: []
+status: In Progress
+assignee:
+  - '@claude'
 created_date: '2026-09-02 21:40'
+updated_date: '2026-09-02 22:07'
 labels: []
 dependencies: []
 ordinal: 316000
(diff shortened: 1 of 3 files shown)


MrLesk:
@codex review

PR #996, opened 2026-09-02
BACK-683 - Render TUI acceptance-criteria progress as a pie glyph

MrLesk:
## Summary

The TUI showed acceptance-criteria progress as an ASCII bar, `[###--] 3/5`. That caution was
aimed at Block Elements, but it was applied to the whole indicator even though the TUI
already renders Geometric Shapes (`● ○ ◒ ✔`) on every supported terminal. The bar also sat
before the task id on In Progress rows only, so ids never lined up across rows.

- One pie glyph plus the live checked/total count replaces the bar: `○` nothing checked,
  `◔` up to a third, `◑` up to two thirds, `◕` above that, `●` only when every criterion is
  checked. It mirrors the ring the web shows and costs one cell.
- Colors are unchanged: green complete, yellow underway, red at a third or fewer. The glyph
  and the color share the same thirds, so `○`/`◔` are red, `◑`/`◕` yellow, `●` green.
- Column reservation moved to `src/ui/task-row-prefix.ts`, which sizes the prefix **per
  render** from the rows actually on screen: a status segment (task list only — board columns
  already name the status) then the progress cell, each padded to the widest of its kind.
  Ids line up, and a render pays only for columns something on it fills.
- The wide/compact variants collapse into one form, taking the `availableWidth` plumbing
  behind them with it.

Which rows show progress is unchanged: In Progress tasks with criteria. Plain and MCP list
output (`(ac: 1/3)`) is untouched.

## Rendered

Task list — custom statuses keep their labels, `100/100` fits, every id starts at one column:

```
lq Tasks (8) qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqk
x                                                           x
x ◒ In Progress ◔ 2/9     TASK-1 - Render progress as a     x
x ◒ In Progress ○ 0/4     TASK-2 - Fail closed on           x
x ◒ In Progress ● 100/100 TASK-3 - Audit every generated    x
x ◒ In Progress           TASK-4 - In progress without      x
x ○ Waiting               TASK-5 - Waiting on the design    x
x ○ Ready                 TASK-6 - Ready to pick up         x
x ○ To Do                 TASK-7 - A plain queued task      x
x ✔ Done                  TASK-8 - A finished task          x
```

Board — the In Progress column reserves for `● 100/100`; the other four columns have no row
with progress, so they spend no prefix width at all:

```
lq ○ To Do (1) qqqqqqqqqqqqqqklq ○ Ready (1) qqqqqqqqqqqqqqklq ◒ In Progress (4) qqqqqqqqklq ○ Waiting (1) qqqqqqqqqqqqklq ✔ Done (1) qqqqqqqqqqqqqqqk
x                            xx                            xx                            xx                            xx                            x
x TASK-7 - A plain queued    xx TASK-6 - Ready to pick up   xx ◔ 2/9     TASK-1 - Render  xx TASK-5 - Waiting on the    xx TASK-8 - A finished task   x
x                            xx                            xx ○ 0/4     TASK-2 - Fail    xx                            xx                            x
x                            xx                            xx ● 100/100 TASK-3 - Audit   xx                            xx                            x
x                            xx                            xx           TASK-4 - In      xx                            xx                            x
```

## Notes for review

**The status word stays.** An earlier revision of this branch reduced the task list to an
icon-only status to buy a fixed-width prefix. `getStatusStyle` maps six statuses, so every
custom status — `Ready`, `Waiting`, `Blocked on review` — collapsed to the same default `○`
and color, making those rows indistinguishable. Per-render sizing gets the alignment without
that trade, so the list shows icon + word on every row again, unconditionally.

**The prefix is emitted before the row-level color tags.** blessed treats a bare `{/}` as a
full attribute reset (`Element.prototype._parseTags`), so the indicator's own close tag used
to cancel the magenta move highlight and the gray cross-branch dim for the rest of the row.
Keeping the prefix outside those tags fixes that, and holds the reserved columns still while
a task moves — the `►` marker now sits between the prefix and the id.

**Board rows carry no status segment.** The column header already names the status and every
row in a column shares it, so a per-row icon would cost a column for no information. Easy to
add if you'd rather have it.

## Width

The five glyphs are East Asian Width "ambiguous", the same class as the `●` already in use.
`unicode.strWidth` in the patched neo-neo-bblessed reports **1** for each, and none matches
the wide-layout regex, so they occupy one cell exactly like the existing status icons. No
Block Elements are introduced. The prefix builder measures with that same `strWidth` rather
than string length, so a non-ASCII status label still pads to the right number of cells.

blessed degrades non-ASCII to `?` only when the locale is not UTF-8 (`Tput.detectUnicode`
reads `LANG`/`LC_ALL`). That already applies to the shipped `◒ ○ ✔` status icons, so the pies
add no new exposure.

## Testing

- `bunx tsc --noEmit`
- `bun run check .`
- `bun run test` — 2860 pass / 8 skip / 1 fail; a clean `origin/main` worktree run gives
  2859 / 8 / 1, failing the same `ContentStore` test. Local-only flake, green on CI.
- PTY QA with `expect` at 150x40 on `board` and `task list`, against a project configured
  with `To Do / Ready / In Progress / Waiting / Done` and a 100-criterion task

### DIFF
--- a/backlog/tasks/back-683 - Render-TUI-acceptance-criteria-progress-as-a-pie-glyph.md
+++ b/backlog/tasks/back-683 - Render-TUI-acceptance-criteria-progress-as-a-pie-glyph.md
@@ -5,7 +5,7 @@ status: In Progress
 assignee:
   - '@claude'
 created_date: '2026-09-02 21:40'
-updated_date: '2026-09-02 22:38'
+updated_date: '2026-09-02 23:00'
 labels: []
 dependencies: []
 ordinal: 315000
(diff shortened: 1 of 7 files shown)


MrLesk:
@codex review

PR #1025, opened 2026-09-17
BACK-688 - Respect most_progressed strategy across branch task copies

### LINKED ISSUE #1024: [Bug]: most_progressed status resolution never runs — workingCopy always wins the tie-break
hey, 
so start with: thanks for this awesome tool! I discovered it yesterday and I already adopted it 🙏

Now to the bug:

## Summary

A task's status on another branch is never reflected in `backlog board` / `backlog task list` / `backlog overview` run from a checkout that already has its own copy of that task file — which is true for essentially every task, since branches are normally created from a commit that already has it. `taskResolutionStrategy: most_progressed` (the default) has no effect in this case: the working copy always wins, regardless of how much further along the branch's status is.

## Repro

```sh
# from any Backlog.md project with git worktrees enabled, on a branch (e.g. master)
# that already has TASK-1 at status "To Do":

git worktree add ../wt-1 -b task-1-work
cd ../wt-1
backlog task edit TASK-1 -s "In Progress"
git add -A && git commit -m "start TASK-1"

cd - # back to the original checkout on master
backlog task list --plain   # TASK-1 still shows "To Do", not "In Progress"
backlog board                # same
```

I confirmed this is not a worktree- or caching-specific issue: I reproduced it with plain `git commit-tree`/`update-ref` against a throwaway local branch, no worktree involved, no long-lived process, fresh CLI invocation each time. Same result. I also confirmed it isn't stale-cache-related (issue #689 / PR #706): every invocation above is a brand-new CLI process with no `contentStore` to go stale.

`config.yml` used: `checkActiveBranches: true`, `remoteOperations: true`, `filesystemOnly: false`, default `taskResolutionStrategy` (`most_progressed`). The progress output confirms the branch is being scanned:

```
Indexing 1 other local branches...
Applying latest task states from branch scans...
```

...but the scanned status never surfaces.

## Root cause

`src/core/task-identity-index.ts`, `selectTaskRecord()` (confirmed against the `v1.52.0` tag):

```ts
function selectTaskRecord(
	records: TaskIdentityRecord[],
	statuses: string[],
	strategy: "most_recent" | "most_progressed",
): TaskIdentityRecord | undefined {
	const candidates = records.filter((record) => record.task);
	return [...candidates].sort((left, right) => {
		if (Boolean(left.workingCopy) !== Boolean(right.workingCopy)) {
			return left.workingCopy ? -1 : 1;
		}
		if (strategy === "most_progressed") {
			const leftRank = Math.max(0, statuses.indexOf(left.task?.status ?? ""));
			const rightRank = Math.max(0, statuses.indexOf(right.task?.status ?? ""));
			if (leftRank !== rightRank) return rightRank - leftRank;
		}
		...
	})[0];
}
```

The `workingCopy` mismatch check runs *before* the strategy-based comparison and returns immediately when it differs. A record built from the current checkout's filesystem (`buildTaskIdentityIndex` in `src/core/backlog.ts`) always sets `workingCopy: true`; a record read from a branch's committed tree (`task-loader.ts`, `BranchTaskLoader.load`) never sets it. So the sort's first comparator always prefers the working copy over any branch record, and the `most_progressed`/`most_recent` logic below it never runs for the case it exists to handle: the current checkout already has some (usually older) copy of the task.

This makes `most_progressed` effectively dead for the common "start a task on a branch, view the board from the branch it was created from" workflow, and I couldn't find an existing issue or PR describing it — the closest, #689, was a different bug (stale corpus in long-lived processes) fixed by #706, whose regression coverage doesn't touch this comparator.

## Suggested fix

Only let `workingCopy` short-circuit the sort when the strategy doesn't already have a more-progressed candidate, or simpler: run the `strategy === "most_progressed"` rank comparison before the `workingCopy` check (or fold `workingCopy` into it as one more input to the rank rather than an unconditional override).

## Environment

- `backlog-md` 1.52.0 (Homebrew), also verified against the `v1.52.0` git tag source
- macOS, Darwin 27.0.0

JanC:
## Summary

Fix task resolution so `most_progressed` compares status rank before the working-copy tie-break. A more-progressed matching task on a scanned branch now appears in list, board, and overview results; equal-progress records still prefer the current checkout.

## Related Issue or Task

Closes #1024

BACK-688 - Respect most_progressed strategy across branch task copies

> **📋 Important:** Please discuss the change in an issue before opening a PR.
> - If no issue exists, open one first so maintainers and contributors can agree that the proposal fits the current project scope before implementation.
> - All PRs must have an associated task in the backlog.
> - If no task exists, create one first using: `backlog task create "Your task title"`
> - Follow the [task guidelines](../src/guidelines/agent-guidelines.md) when creating tasks
> - Tasks should be atomic, testable, and well-defined with clear acceptance criteria

## Task Checklist

## Testing

- `bun test --timeout=10000 src/test/task-identity-index.test.ts` (7 pass)
- `bunx tsc --noEmit`
- `bun run check .`
- `bun run test` was also run. Existing Git hook failures in `src/test/tui-task-composer.test.ts` remain unrelated to this change.


JanC:
I see only now that a PR should be opened only after the discussin. Sorry for that, I'll move this PR back to draft

PR #1028, opened 2026-09-17
Add incremental documentation flags to task edit

### LINKED ISSUE #1027: [Feature]: Add incremental documentation flags to task edit
## Summary

`backlog task edit --doc` replaces the entire `documentation` array. The CLI has no way to add or remove a single entry while preserving the others.

This is the same surface drift #856 reported for references: the shared task-update model already implements `addDocumentation` / `removeDocumentation` (`src/types/task-edit-args.ts`, `src/utils/task-edit-builder.ts` lines 126-133, mapped into `TaskUpdateInput` exactly like references), and the MCP `task_edit` schema exposes them. Only the canonical CLI cannot perform the incremental mutation.

**This is not a new proposal — it is the follow-up the project already recorded and never executed.** BACK-597, the task that shipped `--add-ref`/`--remove-ref` in #870, scoped documentation out on purpose and said so twice:

> Scope: references only. Documentation has the identical drift (MCP exposes addDocumentation/removeDocumentation with no CLI equivalent); record it as a follow-up observation, do not implement it here.

> Follow-up observation (scope guard, not implemented): documentation has the identical drift. MCP task_edit exposes addDocumentation/removeDocumentation and Core.updateTask implements them, but the CLI still has only `--doc` (replace) and `--clear-docs`. A follow-up would add `--add-doc`/`--remove-doc` through the same helpers.

No such follow-up task or issue exists.

## Reproduction

Backlog.md 1.51.0 (verified identical on 1.52.0, current on npm):

```console
$ backlog task create "Doc probe" --doc docs/a.md --doc docs/b.md --plain
Task TASK-1 - Doc probe
Documentation: docs/a.md, docs/b.md

$ backlog task edit task-1 --doc docs/c.md --plain
Documentation: docs/c.md

$ backlog task task-1 --json | jq .task.documentation
[
  "docs/c.md"
]

$ backlog task edit --help | grep -E -- '--(add|remove)-doc'
# no output; grep exits 1
```

`main` at the time of writing has no `--add-doc`/`--remove-doc` and never assigns `editArgs.addDocumentation` in `src/cli.ts`.

## Why it bites in practice

Renaming a documentation page is the trigger. The natural gesture — "repoint the one path that changed" — destroys every other entry, and nothing warns you:

```console
$ backlog task edit 12 --doc <renamed-path> --doc <other-path-I-happened-to-remember>
Updated task TASK-012
```

That command ran against a task carrying six documentation entries and left two. The four dropped entries were exactly the surfaces the task existed to review. Exit code 0, success message, no diff shown. It was not caught by the author; it was caught by an automated code review on the resulting pull request, and a second task in the same repository had been silently truncated the same way.

The read-modify-write workaround also has the concurrency hazard #856 described: a writer that changes `documentation` between the client-side read and the replacing edit is silently discarded. That is a different boundary from the `task edit` locking work in #843, which cannot make a preceding client-side read part of the same critical section.

Worth noting that the flag's own help text is what makes the trap quiet: `--doc <documentation>  set documentation (can be used multiple times)`. "Can be used multiple times" reads as "repeat it to add entries" — which is true within a single invocation, and false across invocations.

## Expected behavior

Expose the already-shipped model operations through `task edit`, alongside the replacement flag:

```console
backlog task edit task-1 --add-doc docs/c.md
backlog task edit task-1 --remove-doc docs/a.md
```

Semantics as shipped: additions preserve existing entries and do not duplicate, removals match by value and leave unrelated entries unchanged. Repeated flags and comma-separated inputs per the existing multi-value convention. `--doc` remains the explicit replace-all operation.

BACK-597 already fixed the conventions this should follow: join `hasEditFieldFlags` so an interactive TTY applies the flags instead of opening the wizard; reuse `validateClearableListInput` per flag so blank values are rejected naming the exact flag; mutual exclusion with `--clear-docs`, and with `--doc` (mirroring the `--label` vs `--add-label`/`--remove-label` rule so replacement and incremental edits stay exclusive); `addHelpSchema` entries; and the Replace/Add/Remove rows in `src/guidelines/agent-guidelines.md`. `src/test/cli-refs-docs.test.ts` already covers both refs and docs, so the reference test cases transpose directly.

## Same drift, not proposed here

`--assignee` replaces too, and has no `addAssignee` in the model — implementing it would be a different, larger change. Flagging it only so the documentation fix is not mistaken for a complete sweep.

## Duplicate search

Searched this repository for `add-doc`, `addDocumentation`, `removeDocumentation` across code and issues. The only two occurrences of `add-doc` in the repository are inside Backlog task markdown files (`back-597`, `back-367`); none in `/src`. No open issue and no open pull request addresses it. #870 mentions `--add-doc` exactly once, in the BACK-597 task file, to declare it out of scope.

Does this fit the current scope? If it does, I'm happy to implement it through the documented workflow — a backlog task created with `backlog task create`, then a PR following the `BACK-xxx` conventions — or to leave it to the maintainers if you'd rather handle it the way BACK-597 handled #856.

BetExchange91:
Closes #1027.

Exposes the already-shipped `addDocumentation` / `removeDocumentation` model operations through `task edit` as `--add-doc` and `--remove-doc` — the follow-up BACK-597 recorded when #870 added `--add-ref` / `--remove-ref`:

> A follow-up would add `--add-doc`/`--remove-doc` through the same helpers.

This PR is that follow-up, and deliberately nothing more: it mirrors #870 point for point (`src/cli.ts` +69/−8 here vs +63/−6 there; the test file +127 in both).

## Changes

**`src/cli.ts`** — no new merge logic; `buildTaskUpdateInput` and `Core.updateTask` already implement both operations.

- `hasEditFieldFlags` includes `addDoc` / `removeDoc`, so an interactive TTY applies the flags instead of opening the edit wizard.
- `validateTaskListFlags` gains one `validateClearableListInput` per new flag: blank values rejected naming the exact flag, conflict with `--clear-docs` rejected.
- `--doc` combined with `--add-doc` / `--remove-doc` is rejected, with the same wording as the `--ref` rule, so replacement and incremental edits stay mutually exclusive.
- `addEditFieldOptions` declares both flags with `createMultiValueAccumulator()` (repeatable and comma-separated). `--doc` moves to the same helper — byte-identical to the inline closure it replaces, as #870 did for `--ref` — and its help text now says "replace all documentation".
- Help schema entries for `task edit` and `draft edit`.

**`src/guidelines/agent-guidelines.md`** — Add / Remove documentation rows next to the existing Replace row.

**`src/test/cli-refs-docs.test.ts`** — the nine `--add-ref` / `--remove-ref` cases, transposed: add preserves and dedups; remove matches by value and leaves references untouched; repeated and comma forms; add+remove of the same value in one command pinned to the shared model order (removal wins); interactive TTY; blank values, `--clear-docs` and `--doc` conflicts exit 1 without mutating; `--doc` still replaces; help lists both flags.

Unchanged: `task create` (no incremental semantics on a new task), MCP `task_edit` behavior, and every existing `--doc` / `--clear-docs` error string.

## Verification

- `bunx tsc --noEmit` — clean.
- `bun run check .` — clean, 431 files.
- `bun test src/test/cli-refs-docs.test.ts` — 46 pass, 0 fail (37 existing + 9 new).
- `bun test` on `mcp-refs-docs`, `documentation`, `references`, `mcp-tasks`, `agent-instructions` — 85 pass, 0 fail.
- Live run in a scratch project from `src/cli.ts`: `a.md, b.md` → `--add-doc c.md` → `a.md, b.md, c.md`; `--add-doc a.md` dedups; `--remove-doc b.md` → `a.md, c.md`; `--add-doc ""`, `--clear-docs --add-doc`, and `--doc … --add-doc …` each exit 1 with their message and leave the list unchanged.

**Full suite, reported as it ran on my machine:** 2888 pass, 8 skip, 8 fail, 1 error. None of the failing tests touch `task edit` or documentation. To separate cause from environment I re-ran the failing files on a clean `upstream/main` worktree (`aded8e2`) without this commit: the same tests fail there — `config-commands` (2), `cli-list-window` (3), `mcp-client-setup` (1), and `server-statistics-endpoint`, which failed on the baseline run and passed on the patched one. `content-store` passes in isolation on both. The failures look environmental on macOS (ANSI color codes in expected error strings, watcher and lock timing), so CI should be the judge here.

Not included: `--assignee` has the same replace-only surface, but the model has no `addAssignee` — implementing it is a separate, larger change, noted in #1027.


BetExchange91:
Closing this: it was opened too early. The PR template asks for the change to be discussed in an issue first, and for every PR to have a backlog task created with `backlog task create` — neither happened, and the branch, commit and title don't follow the `BACK-xxx` conventions either. Apologies for the noise.

The proposal stays open for discussion in #1027. If it fits the project's scope, I'll follow up with a PR that goes through the documented workflow.

